PR #1, opened 2020-02-09
[WIP] Coding Style improved and few potential errors fixed

Spomky:
This PR tries to fix some potential bugs detected by Psalm.
It also changes the PHP-CS-Fixer configuration by adding new rules.

Changes:
* Psalm added to spot potential errors. Level reached at the time of writing: 6
* New PHP-CS-Fixer configuration with more rules.
* New Travis-CI configuration
  * Matrix is kept
  * Adds PHP 8 test (even if failing for now)
  * Adds coding standard and quality checks
  * Adds security checks


src/bundle/Controller/FormController.php
```diff
@@ -96,7 +105,7 @@ protected function getTemplateVars(Request $request, TwoFactorTokenInterface $to
     {
         $config = $this->twoFactorFirewallContext->getFirewallConfig($token->getProviderKey());
         $pendingTwoFactorProviders = $token->getTwoFactorProviders();
-        $displayTrustedOption = $this->trustedFeatureEnabled && (!$config->isMultiFactor() || 1 === count($pendingTwoFactorProviders));
+        $displayTrustedOption = $this->trustedFeatureEnabled && (!$config->isMultiFactor() || 1 === \count($pendingTwoFactorProviders));
```
scheb: I have to say, until today I've never understoods what's the thing with with prefixing built-in global functions with `\`. Could you explain to me?

Btw. further down there are some function calls without a `\` prefix, which confuses me even more when `\` is added and when it isn't. Cannot make out a pattern that makes sense to me.

src/bundle/Security/Http/Authentication/DefaultAuthenticationRequiredHandler.php
```diff
@@ -46,7 +55,7 @@ public function onAuthenticationRequired(Request $request, TokenInterface $token
     {
         // Do not save the target path when the current one is the one for checking the authentication code. Then it's
         // another redirect which happens in multi-factor scenarios.
-        if (!$this->isCheckAuthCodeRequest($request) && $request->hasSession() && $request->isMethodSafe(false) && !$request->isXmlHttpRequest()) {
+        if (!$this->isCheckAuthCodeRequest($request) && $request->hasSession() && $request->isMethodSafe() && !$request->isXmlHttpRequest()) {
```
scheb: Missed this when cleaning up for 4.4, will take it over as a clean-up change.

src/bundle/Security/TwoFactor/Event/Event.php
```diff
@@ -2,19 +2,19 @@
...
 
-if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class)) {
-    class Event extends ContractsBaseEvent
-    {
-    }
-} else {
-    class Event extends BaseEvent
-    {
-    }
+class Event extends ContractsBaseEvent
```
scheb: Missed this when cleaning up for 4.4, will take it over as a clean-up change.

psalm.xml
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0"?>
```
scheb: I've never worked with psalm before, so I'm wondering where's this config coming from? Is that like a recommended set of checks?

composer.json
```diff
@@ -29,7 +29,8 @@
     "require-dev": {
         "doctrine/lexer": "^1.0.1",
         "doctrine/orm": "^2.6",
-        "phpunit/phpunit": "^7.0|^8.0",
+        "egulias/email-validator": "^2.1.10",
```
scheb: Where's this being used? 🤔

.travis.yml
```diff
@@ -1,37 +1,82 @@
...
+  - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
 
 install:
   - travis_retry composer update --prefer-dist --no-interaction --no-suggest $COMPOSER_FLAGS
 
 script:
-  - mkdir -p build/logs
-  - ./vendor/bin/phpunit --coverage-clover build/logs/clover.xml
-  - ./vendor/bin/phpcs --standard=php_cs.xml --ignore=*/vendor/* .
```
scheb: Err, where did the code style go? 🤔 

Did you replace it with a run of php-cs-fixer? If that's the case, I have two concerns about this approach:

1) php-cs-fixer is not a linter tool, so this solution doesn't allow running the code style linter in the IDE to display violations
2) I've recently added a code style rule to enforce constant visibility and that check is gone now, since php-cs-fixer only cares about formatting issues (maybe psalm has a rule for this?)

src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
```diff
@@ -15,6 +24,9 @@
  */
 class AuthenticationProviderDecoratorCompilerPass implements CompilerPassInterface
 {
+    /**
```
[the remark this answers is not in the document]
jdreesen: That's correct, but technically it would be a BC break here because the class is not final, so someone might extend it (even if it's unlikely in this case).

src/bundle/Controller/FormController.php
jdreesen: 1. When a global function is called from inside in namespace, it has a tiny little overhead (only for the first call if OPCache is enabled, and even then not really measurable) because PHP has to check if there's a function with the same name inside the current namespace before calling the global one.

2. There are some "special" core functions with compiler optimized code that only gets triggered when explicitly called as global function. The difference for these *is* measurable, so often only these get prefixed. But it's hard to remember which they are, so it's best to let a tool like PHP-CS-Fixer or [Php Inspections ​(EA Extended)](https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-) (or even [Ultimate](https://plugins.jetbrains.com/plugin/10215-php-inspections-ea-ultimate-)) do this for you.<br>
A list of these functions can be found here, for example (contains links to the PHP sources, too): https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/9b3716c623bde9f497022675acef31a60fa19561/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L357-L397

Prefixing the functions with `\` or using them via `use function ...` makes no difference, it's just a matter of taste.

src/backup-code/Model/BackupCodeInterface.php
```diff
@@ -1,5 +1,16 @@
 <?php
 
+declare(strict_types=1);
+
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Christian Scheb
```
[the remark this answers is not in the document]
Spomky: OK no problem. Indeed it is generated automatically for all files by PHP-CS-Fixer so the maintenance is not so difficult.
I will revert that.

src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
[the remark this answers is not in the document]
Spomky: I had some trouble extending classes or implementing interfaces when the method signature is not identical the the parent one. TBD

src/bundle/Controller/FormController.php
Spomky: As @jdreesen said, that's a micro optimization when calling this function from a namespace.
I will revert it to avoid confusion

psalm.xml
Spomky: This file is the default configuration when you run `psalm --init src/ 8`. With level 8, most of rules are marked as information, not errors.
Then I progessively raised the level by removing those info rules.

composer.json
Spomky: It is used by `swiftmailer/swiftmailer` and I had a bug when testing with ``--prefer-lowest`.
By the way, why `swiftmailer/swiftmailer` in the dev reqs? Is it possible to review the associated tests and mock what should be mocked?

.travis.yml
Spomky: OK so I'll keep it. I was not sure for that change.

Spomky:
Thank you for the review.
I will update this PR soon.

src/bundle/Controller/FormController.php
scheb: Ok, I just wanted to understand. So let's keep it, but then we should make sure we have this consistent within the codebase.

src/bundle/Security/TwoFactor/Event/Event.php
scheb: Added to https://github.com/scheb/2fa/commit/dd1d48194488b009ff482674021350b33d1487c6

src/bundle/Security/Http/Authentication/DefaultAuthenticationRequiredHandler.php
scheb: Added to https://github.com/scheb/2fa/commit/dd1d48194488b009ff482674021350b33d1487c6

composer.json
scheb: It's used here https://github.com/scheb/2fa/blob/master/src/email/Mailer/AuthCodeMailer.php#L26 and in the related test case.

But since we don't want to force people into using that specific mailer it's not part of `require` but only `require-dev` to make it available to tests.

src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
```diff
@@ -40,6 +52,6 @@ private function decorateAuthenticationProvider(ContainerBuilder $container, str
 
     private function isTwoFactorProvider(string $authenticationProviderId): bool
     {
-        return false !== strpos($authenticationProviderId, TwoFactorFactory::PROVIDER_ID_PREFIX);
+        return false !== mb_strpos($authenticationProviderId, TwoFactorFactory::PROVIDER_ID_PREFIX);
```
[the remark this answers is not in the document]
jdreesen: You're probably right. I assume the code has been changed through a PHP-CS-Fixer setting.

scheb:
I took it over and just merged into master. Thank you for the good suggestions! Helped a lot to bring the CI and code quality to the next level.

Might back-port this scheb/two-factor-bundle.

PR #3, opened 2020-02-12
Added support for Symfony Mailer component

philomather:


tests/Mailer/SymfonyAuthCodeMailerTest.php
```diff
@@ -0,0 +1,72 @@
...
+    private $symfonyMailer;
+
+    /**
+     * @var SymfonyAuthCodeMailer
+     */
+    private $mailer;
+
+    protected function setUp(): void
+    {
+        if (\interface_exists(MailerInterface::class) === false) {
```
scheb: I believe this can add the mailer component to composer.json `require-dev` now, since it's available for any Symfony version this version of the bundle supports. And then this check can be removed because we can execute the test in all test configurations.

philomather: Yes that seems to work!

scheb:
Looks good, thank your for your work and moving the changes over to this repository 👍

PR #8, opened 2020-05-13
Experimental authenticator integration

### LINKED ISSUE #5: Integrate with new "Authenticators" security
Integrate the bundle with the new "authenticator" security system, which was introduced as an experimental feature in Symfony 5.1.

The old approach doesn't work anymore, which was: retrieving all "authentication listeners" from `security.authentication.manager` to decorate them and effectively wrap away the authenticated token before it becomes visible to the security system. There are no more listeners that could be used as a hook-in point

Potentially the new concept of "badges" and event-based authentication can help. Could potentially lead the bundle towards a cleaner, less hacky integration. Currently unclear, needs more investigation.

References:

- https://github.com/symfony/symfony/pull/33558
- https://wouterj.nl/2020/04/authenticators-new-symfony-security

Notes:

Token is created by the authenticator:
https://github.com/wouterj/symfony/blob/b1e040f311e16f4888f42564a6d5da0a489c929a/src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php#L169

Example:
https://github.com/wouterj/symfony/blob/b1e040f311e16f4888f42564a6d5da0a489c929a/src/Symfony/Component/Security/Http/Authenticator/AbstractAuthenticator.php#L35

scheb:
Closes #5

### DIFF
--- a/src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
+++ b/src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
@@ -8,6 +8,7 @@ use Scheb\TwoFactorBundle\DependencyInjection\Factory\Security\TwoFactorFactory;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\OutOfBoundsException;
 use Symfony\Component\DependencyInjection\Reference;
 
 /**
@@ -17,15 +18,29 @@ class AuthenticationProviderDecoratorCompilerPass implements CompilerPassInterfa
 {
     public function process(ContainerBuilder $container): void
     {
-        $authenticationManager = $container->getDefinition('security.authentication.manager');
-        $authenticationProviderIds = $authenticationManager->getArgument(0)->getValues();
-        /** @var Reference $authenticationProvider */
-        foreach ($authenticationProviderIds as $authenticationProvider) {
+        foreach ($this->getAuthenticationProviderIds($container) as $authenticationProvider) {
             // Ensure not to decorate the two-factor authentication provider, otherwise we'll get an endless loop
             $authenticationProviderId = (string) $authenticationProvider;
             if (!$this->isTwoFactorProvider($authenticationProviderId)) {
                 $this->decorateAuthenticationProvider($container, $authenticationProviderId);
             }
+        }
+    }
+
+    /**
+     * @return Reference[]
+     */
+    private function getAuthenticationProviderIds(ContainerBuilder $container): array
+    {
+        if (!$container->hasDefinition('security.authentication.manager')) {
+            return [];
+        }
+
+        try {
+            return $container->getDefinition('security.authentication.manager')->getArgument(0)->getValues();
+        } catch (OutOfBoundsException $e) {
+            // Authentication providers are not set, so the authenticator system is used
+            return [];
         }
     }
 
--- a/src/bundle/DependencyInjection/Compiler/RememberMeServicesDecoratorCompilerPass.php
+++ b/src/bundle/DependencyInjection/Compiler/RememberMeServicesDecoratorCompilerPass.php
@@ -16,22 +16,29 @@ use Symfony\Component\DependencyInjection\Reference;
  */
 class RememberMeServicesDecoratorCompilerPass implements CompilerPassInterface
 {
-    private const REMEMBER_ME_LISTENER_ID_PREFIX = 'security.authentication.listener.rememberme.';
+    private const REMEMBER_ME_AUTHENTICATION_LISTENER_ID_PREFIX = 'security.authentication.listener.rememberme.';
+    private const REMEMBER_ME_AUTHENTICATOR_LISTENER_ID_PREFIX = 'security.listener.remember_me.';
 
     public function process(ContainerBuilder $container): void
     {
         // Find all remember-me listener definitions
         foreach ($container->getDefinitions() as $definitionId => $definition) {
-            if (0 === strpos($definitionId, self::REMEMBER_ME_LISTENER_ID_PREFIX)) {
-                $this->decorateRememberMeServices($container, $definition);
+            // Classic security system
+            if (0 === strpos($definitionId, self::REMEMBER_ME_AUTHENTICATION_LISTENER_ID_PREFIX)) {
+                $this->decorateRememberMeServices($container, $definition, 1);
+            }
+
+            // Authenticator security system
+            if (0 === strpos($definitionId, self::REMEMBER_ME_AUTHENTICATOR_LISTENER_ID_PREFIX)) {
+                $this->decorateRememberMeServices($container, $definition, 0);
             }
         }
     }
 
-    private function decorateRememberMeServices(ContainerBuilder $container, Definition $authListenerDefinition): void
+    private function decorateRememberMeServices(ContainerBuilder $container, Definition $authListenerDefinition, int $argument): void
     {
         // Get the remember-me services from the listener and decorate it
-        $rememberMeServicesId = (string) $authListenerDefinition->getArgument(1);
+        $rememberMeServicesId = (string) $authListenerDefinition->getArgument($argument);
         if ($rememberMeServicesId) {
             $decoratedServiceId = $rememberMeServicesId.'.two_factor_decorator';
             $container
--- a/src/bundle/DependencyInjection/Factory/Security/TwoFactorFactory.php
+++ b/src/bundle/DependencyInjection/Factory/Security/TwoFactorFactory.php
@@ -29,6 +29,7 @@ class TwoFactorFactory implements SecurityFactoryInterface
     public const DEFAULT_CSRF_PARAMETER = '_csrf_token';
     public const DEFAULT_CSRF_TOKEN_ID = 'two_factor';
 
+    public const AUTHENTICATOR_ID_PREFIX = 'security.authenticator.two_factor.';
     public const PROVIDER_ID_PREFIX = 'security.authentication.provider.two_factor.';
     public const LISTENER_ID_PREFIX = 'security.authentication.listener.two_factor.';
     public const SUCCESS_HANDLER_ID_PREFIX = 'security.authentication.success_handler.two_factor.';
@@ -39,6 +40,7 @@ class TwoFactorFactory implements SecurityFactoryInterface
     public const KERNEL_EXCEPTION_LISTENER_ID_PREFIX = 'security.authentication.kernel_exception_listener.two_factor.';
     public const KERNEL_ACCESS_LISTENER_ID_PREFIX = 'security.authentication.access_listener.two_factor.';
 
+    public const AUTHENTICATOR_DEFINITION_ID = 'scheb_two_factor.security.authenticator';
     public const PROVIDER_DEFINITION_ID = 'scheb_two_factor.security.authentication.provider';
     public const LISTENER_DEFINITION_ID = 'scheb_two_factor.security.authentication.listener';
     public const SUCCESS_HANDLER_DEFINITION_ID = 'scheb_two_factor.security.authentication.success_handler';
@@ -48,6 +50,16 @@ class TwoFactorFactory implements SecurityFactoryInterface
     public const PROVIDER_PREPARATION_LISTENER_DEFINITION_ID = 'scheb_two_factor.security.provider_preparation_listener';
     public const KERNEL_EXCEPTION_LISTENER_DEFINITION_ID = 'scheb_two_factor.security.kernel_exception_listener';
     public const KERNEL_ACCESS_LISTENER_DEFINITION_ID = 'scheb_two_factor.security.access_listener';
+
+    /**
+     * @var TwoFactorServicesFactory
+     */
+    private $twoFactorServicesFactory;
+
+    public function __construct(TwoFactorServicesFactory $twoFactorServicesFactory)
+    {
+        $this->twoFactorServicesFactory = $twoFactorServicesFactory;
+    }
 
     public function addConfiguration(NodeDefinition $node): void
     {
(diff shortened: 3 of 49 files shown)


src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
```diff
@@ -17,6 +17,12 @@ class AuthenticationProviderDecoratorCompilerPass implements CompilerPassInterfa
 {
     public function process(ContainerBuilder $container): void
     {
+        if ($container->hasDefinition('security.authenticator.manager')) {
+            // Skip integration with the old security system when the new authenticator manager is present
+            // Any better way to do this?
+            return;
+        }
+
```
scheb, line 25: What's the best way for me as as 3rd party bundle to find out if the authenticator or the old system is used? I don't have access for the security bundle configuration, or do I?

For compliler passes I can check for `hasDefinition('security.authenticator.manager')`, since they're applied after the security bundle was loaded.

Unfortunately this doesn't work in
- `DependencyInjection/Factory/Security/TwoFactorFactory.php`
- `DependencyInjection\SchebTwoFactorExtension.php`
- `SchebTwoFactorBundle.php`

because when the bundle loads, the service definitions from the security bundle haven't necessary loaded yet.

src/bundle/DependencyInjection/Compiler/RememberMeServicesDecoratorCompilerPass.php
```diff
@@ -16,22 +16,28 @@
...
         // Find all remember-me listener definitions
         foreach ($container->getDefinitions() as $definitionId => $definition) {
-            if (0 === strpos($definitionId, self::REMEMBER_ME_LISTENER_ID_PREFIX)) {
-                $this->decorateRememberMeServices($container, $definition);
+            if (0 === strpos($definitionId, self::REMEMBER_ME_AUTHENTICATION_LISTENER_ID_PREFIX)) {
+                $this->decorateRememberMeServices($container, $definition, 1);
+            }
+
+            // New security authenticator system
+            if (0 === strpos($definitionId, self::REMEMBER_ME_AUTHENTICATOR_LISTENER_ID_PREFIX)) {
```
scheb, line 31: This is quite ugly. With authenticators I could also decorate the listener itself to prevent the remember-me cookie from being set.

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
```diff
@@ -0,0 +1,228 @@
...
+            return false;
+        }
+
+        $currentToken = $this->tokenStorage->getToken();
+//        return $this->isCheckAuthCodeRequest($request)
+//            || $this->isAuthFormRequest($request)
+//            // Let routes pass, e.g. if a route needs to be callable during two-factor authentication
+//            || !$this->twoFactorAccessDecider->isAccessible($request, $currentToken);
+
+        return $currentToken instanceof TwoFactorTokenInterface && $currentToken->getProviderKey() === $this->firewallName;
```
scheb, line 129: Is this `$currentToken->getProviderKey() === $this->firewallName` necessary or does the new system guarantee that only authenticators from the firewall itself are called?

scheb:
I need to redirect the user to the 2fa form, while they're in that intermediate "2fa not finished" state. Until now, I've done this in the authentication listener:

https://github.com/scheb/2fa/blob/5d25f887d62353bd6c10b66b1116aaf7e99920cf/src/bundle/Security/Http/Firewall/TwoFactorListener.php#L173-L175

With authenticators I can't do this anymore, because authenticators can only generate a response when the login was successful/failed and not on every request.

**Would it be feasible to have my own `kernel.request` listener for that, which is called directly after the firewall and generates a redirect response?**

src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
wouterj: Checking for the `security.authenticator.manager` service definition in a compiler pass is imho the correct solution. For the other cases:

* The `createAuthenticator()` method is only called when using the authenticator and the `create()` method is only called when using the old authenticator. So there is no need for checking it in these methods.
* I don't see you using it in the extension atm. If you need it, I would say: populate the container with both the old and new system. As the services are private, they will be removed when not used (and otherwise, you can use a compiler pass to remove/tweak the definitions).
* The bundle should not need to know it as well, the compiler pass can check for the service ID and return early if the old system is used.

src/bundle/Security/Http/Authenticator/AuthenticatorDecorator.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return $this->decoratedAuthenticator->authenticate($request);
+    }
+
+    public function createAuthenticatedToken(PassportInterface $passport, string $firewallName): TokenInterface
+    {
+        $token = $this->decoratedAuthenticator->createAuthenticatedToken($passport, $firewallName);
+
+        // AnonymousToken and TwoFactorTokenInterface can be ignored
+        if ($token instanceof AnonymousToken || $token instanceof TwoFactorTokenInterface) {
```
wouterj, line 75: As this is the new system, `AnonymousToken` can be removed.

src/bundle/Security/Http/Authenticator/AuthenticatorDecorator.php
```diff
@@ -0,0 +1,119 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\AnonymousToken;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Http\Authenticator\AuthenticatorInterface;
+use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
+
+class AuthenticatorDecorator implements AuthenticatorInterface
```
wouterj, line 19: Please implement `InteractiveAuthenticatorInterface` as well and then add:

```php
public function isInteractive(): bool
{
    return $this->decoratedAuthenticator instanceof InteractiveAuthenticatorInterface && $this->decoratedAuthentication->isInteractive();
}
```

That would make this authenticator fully compatible for all authenticators (I don't want to tell much about the `InteractiveAuthenticatorInterface`, as it's probably going to be deprecated/removed along with the `INTERACTIVE_LOGIN` event in an upcoming version)

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
wouterj: Yes, it does (the new system has one `AuthenticatorManager` instance per firewall, so it's only aware of the authenticators for that specific firewall)

wouterj:
> Would it be feasible to have my own kernel.request listener for that, which is called directly after the firewall and generates a redirect response?

From what I understand, this code is executed when all authenticators failed to authenticate the request? Would listening on the new `LoginSuccessEvent` or `LoginFailedEvent` (or both?) work?

src/bundle/DependencyInjection/Compiler/AuthenticationProviderDecoratorCompilerPass.php
scheb: > I don't see you using it in the extension atm. If you need it, I would say: populate the container with both the old and new system. As the services are private, they will be removed when not used (and otherwise, you can use a compiler pass to remove/tweak the definitions).

I thought I can skip some unnecessary service definitions ;). But I'm also fine with just loading them, even when they're not used.

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
scheb: Great, another parameter that I can remove from the constructor :)

scheb:
@wouterj Thank you for the feedback, very helpful!

> Did a very quick scan (I'll test somewhere next week, I'm very busy till tuesday), looks great afaics!

If you want to play around with it, the repository contains a little test application located in `app`. Just run `composer install` and `symfony serve` in that directory to start it up.

> From what I understand, this code is executed when all authenticators failed to authenticate the request? Would listening on the new LoginSuccessEvent or LoginFailedEvent (or both?) work?

No, the code needs to be executed as long as the user is in that intermediate state, when 2fa has not been completed yet. The purpose is to redirect the user to the 2fa form page, to ask for the 2fa code. The events you mentioned would only trigger when an authenticator was executed within the request.

I was playing around with different solutions today and I'll most likely add a listener for `kernel.request` to detect the "2fa incomplete" state and redirect the user. Alternatively, I'll investigate to add my own `AccessListener`-like implementation, which would handle the "2fa not completed" state.

I can give you an update as soon as I have something running that I believe is feasible.

wouterj:
Ah, please note that the firewall listeners (like `AccessListener` and `ContextListener`) have not been removed. So I can also think of a solution where we create a new `ListenerFactoryInterface` that would allow custom factories to create firewall listeners.

That would allow you to reuse the lazy firewall logic, which I think would make sense.

scheb:
Yea, would be cool to have an easier way to inject a listener into the firewall. Having  a `ListenerFactoryInterface` sounds like a good solution to me. For now I'll probably go for a compiler pass, to add it to the list of listeners. I have some unfinished code laying around, but I need to do some other refactorings first.

This whole authenticator experiment uncovered some bad design choices in my bundle and I figured there's better ways to do it. So I'm working on that at the moment ;)

src/bundle/Security/Http/Authenticator/Passport/TwoFactorPassport.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function checkIfCompletelyResolved(): void
+    {
+        foreach ($this->badges as $badge) {
+            if (!$badge->isResolved()) {
+                throw new InvalidTwoFactorCodeException('Invalid two-factor authentication code.');
+            }
+        }
+    }
```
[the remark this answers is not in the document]
scheb: I wasn't sure if `PassportTrait` would be safe to use for a 3rd party. I want to avoid behavioural changes being silently introduced, so I decided to write these few lines on my on. What's your view on this?

src/bundle/Security/Http/Authenticator/Passport/TwoFactorPassport.php
```diff
@@ -0,0 +1,64 @@
...
+    public function getBadge(string $badgeFqcn): ?BadgeInterface
+    {
+        return $this->badges[$badgeFqcn] ?? null;
+    }
+
+    public function checkIfCompletelyResolved(): void
+    {
+        foreach ($this->badges as $badge) {
+            if (!$badge->isResolved()) {
+                throw new InvalidTwoFactorCodeException('Invalid two-factor authentication code.');
```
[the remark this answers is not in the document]
scheb: Good point, haven't thought about that.

I'm wondering, how would you distinguish between different error types then? Looks to me as if in the Symfony code everything becomes a `BadCredentialsException`, so how would you tell if it's wrong credentials vs. CSRF error?

src/bundle/DependencyInjection/Factory/Security/TwoFactorFactory.php
```diff
@@ -110,6 +122,46 @@ public function create(ContainerBuilder $container, $id, $config, $userProvider,
...
+        $authenticatorId = self::AUTHENTICATOR_ID_PREFIX.$firewallName;
+        $container
+            ->setDefinition($authenticatorId, new ChildDefinition(self::AUTHENTICATOR_DEFINITION_ID))
+            ->replaceArgument(0, new Reference($twoFactorFirewallConfigId))
+            ->replaceArgument(2, new Reference($successHandlerId))
+            ->replaceArgument(3, new Reference($failureHandlerId))
+            ->replaceArgument(4, new Reference($authRequiredHandlerId));
+
+        return $authenticatorId;
+    }
```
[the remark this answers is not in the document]
scheb: Technically, it could. Conceptually, I thought of `AuthenticatorTwoFactorFactory` more as the construct I use to create compatibility between different Symfony versions, rather than an actual class.

scheb:
Thanks for looking into it. As you can see from the many force-pushes, I'm still messing around a lot in the code, but I consider the implementation now relatively stable.

Good that you mentioned `PUBLIC_ACCESS`, I totally forgot about this new attribute. Had a look at https://github.com/wouterj/2fa/commit/40699ef2424ed32ec7994e9e154e1570bd696643 and this looks good to me. Now I'm glad that I didn't remove that class, something I was thinking about recently 😆 I'd also need to backport the change to `scheb/two-factor-bundle` for full compatibility with Symfony 5.1.

src/bundle/Security/Http/Authenticator/Passport/TwoFactorPassport.php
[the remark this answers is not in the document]
wouterj: I intentionally didn't add `@experimental` to the trait, so it is included in the BC promise (once out of experimental state).

The way I see it is exactly the opposite: For any non-2fa related things in the passport, you want it to behave as it does for the Symfony version it's used in. Let's say Symfony changes things in the passport in 5.2. If you use the trait, Symfony would have updated it as well and the current `TwoFactorPassport` is already compatible. If you don't use the trait, you would have to update your class (and maybe even do some BC magic).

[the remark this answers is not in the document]
scheb: Thanks for the clarification, using `PassportTrait` now.

src/bundle/Security/Http/Authenticator/Passport/TwoFactorPassport.php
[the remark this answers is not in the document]
scheb: Refactored so that this exception is throw in the `CheckTwoFactorCodeListener` now.

scheb:
Almost done, some test cases are still missing :)

wouterj:
Very cool to see this merged. 🎈 Thanks for your hard work! It's one of the first major security bundles with support in their master branch

scheb:
Thanks for your help to make this work! I'll release a v5 with authenticators support soon.

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
```diff
@@ -0,0 +1,183 @@
...
+    private $eventDispatcher;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+    /**
+     * @var TwoFactorAccessDecider
+     */
+    private $twoFactorAccessDecider;
```
Spomky, line 71: Hello @scheb,

I see you inject a `TwoFactorAccessDecider` object, but it is not used in this class.

scheb: Oh, that must have slipped through in the many refactorings. Thanks for noting that, have removed that unused dependency.

PR #19, opened 2020-07-16
Allow to pass variables when creating instance.

Warxcell:
I'm creating custom provider and I need to pass my own variables. Currently I'm extending this class. This would be nice to be in core. :)

### DIFF
--- a/src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
+++ b/src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
@@ -20,7 +20,7 @@ class DefaultTwoFactorFormRenderer implements TwoFactorFormRendererInterface
      */
     private $template;
 
-    private $context
+    private array $context;
 
     public function __construct(Environment $twigRenderer, string $template, array $context = [])
     {
--- a/src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
+++ b/src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
@@ -20,15 +20,18 @@ class DefaultTwoFactorFormRenderer implements TwoFactorFormRendererInterface
      */
     private $template;
 
-    public function __construct(Environment $twigRenderer, string $template)
+    private $context
+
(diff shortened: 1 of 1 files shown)


src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
```diff
@@ -20,15 +20,18 @@ class DefaultTwoFactorFormRenderer implements TwoFactorFormRendererInterface
...
     {
         $this->template = $template;
         $this->twigEnvironment = $twigRenderer;
+        $this->context = $context;
     }
 
     public function renderForm(Request $request, array $templateVars): Response
     {
-        $content = $this->twigEnvironment->render($this->template, $templateVars);
+        $content = $this->twigEnvironment->render($this->template, array_merge($this->context, $templateVars));
```
scheb, line 34: Please add a test case to ensure the variables passed in the constructur are passed to the template.

src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
```diff
@@ -20,15 +20,18 @@ class DefaultTwoFactorFormRenderer implements TwoFactorFormRendererInterface
      */
     private $template;
 
-    public function __construct(Environment $twigRenderer, string $template)
+    private array $context;
```
scheb, line 23: 1) Please annotate this with PHPDoc type `array`
2)  `$context` doesn't make clear what it is. Let's name this `$templateVars`. Please also update the constructur argument.

scheb (changes requested):

src/bundle/Security/TwoFactor/Provider/DefaultTwoFactorFormRenderer.php
scheb: Please use PHPDoc instead of PHP 7.4 type hints. The library has to support PHP versions 7.2 and 7.3, which don't have that language feature.

scheb:
@Warxcell Do you still need this feature? Will you do the changes?

scheb:
I'm taking this one over -> 8a9315b8f169859c1fdbfed37a828fa5ff36a3b8

Issue #21, opened 2020-07-31
Trusted Devices do not work with Symfony 4


hco:
**Bundle version**: 5.0.0
**Symfony version**: 5.1.1

**Description**
If I see it correctly, the trusted device libraries does not work with Symfony before 5, because it depends on the new LoginSuccessEvent and Passports.

I think it would be great to mention this in the compatibility or make it work with the "old" security system as well.

**Additional Context**
Thank you very much for your work on this library, very much appreciated. It works quite well for me even though I am using it in an REST API, which is outside of the scope of your project, if I understand it correctly :-)

scheb:
Are you sure about this? I mean, have you been able to reproduce this issue with sf4.4? I'm asking because it's running the integrating test suite with sf4.4 (https://github.com/scheb/2fa/blob/5.x/.travis.yml#L49:L53) and everything is fine there. The bundle should support the exact same feature set on both authenticator and the old system.

Glad you find the bundle useful. Talking about the API implementation, I'd like to learn how you implemented/configured it. It's over of the most frequent questions I get how to use the bundle on an API and since I never did it myself I can't give a definitive answer. Ideally I'd like to add some info to the docs.

hco:
Not 100% sure, but it doesn't work for me and my feeling while debugging it was that it was related to this.

I'll continue debugging later this week, and keep you posted.

One remark: I am using a custom guard authenticator, this might be related.

Regarding the REST-thingy: I'll share a bit about what I did later. I was already hoping I could contribute parts of it, but I think it's mostly configuration, so I guess we only need to create some docs.

scheb:
To help you debugging, with the old security system the trusted device cookie would be set here:

https://github.com/scheb/2fa/blob/5.x/src/bundle/Security/Http/Firewall/TwoFactorListener.php#L181-L187

and by-passing the 2fa when the device is flagged as trusted would be handled here:

https://github.com/scheb/2fa/blob/5.x/src/trusted-device/Security/TwoFactor/Handler/TrustedDeviceHandler.php#L44-L52

hco:
Thank you, very much appreciate your support!

hco:
Thank you for the links. You are right, it is actually working. I missed the `_trusted` parameter :-(

One of the downsides of using this in a non-documented way ;-)

hco:
[...]

PR #24, opened 2020-08-25
Documentation on how to implement 2fa in an API

scheb:
Adding some instructions how to setup the bundle in an API application, that doesn't render a frontend directly.

Following up on https://github.com/scheb/2fa/issues/21#issuecomment-669262867


scheb:
@hco Any comments/additions? Feel free to just comment in the diff. Thanks.

PR #41, opened 2020-12-04
Change composer install to composer require

ThomasTr:


scheb:
True, that should be composer require. Thanks for fixing that!

PR #43, opened 2020-12-24
Clean up preparation listener state

mathieudz:
Needed for usage of 2fa in combination with PHP PM. As the listener service is reused, the state must be cleared, otherwise it leaks into the next request.
More specifically, in the next request the user is completely authenticated, but the listener would think that the 2fa process is still ongoing.

### DIFF
--- a/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
+++ b/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
@@ -109,16 +109,20 @@ class TwoFactorProviderPreparationListener implements EventSubscriberInterface
         if (!$event->isMasterRequest()) {
             return;
         }
-        if (!($this->twoFactorToken instanceof TwoFactorTokenInterface)) {
+
+        $twoFactorToken = $this->twoFactorToken;
+        $this->twoFactorToken = null;
+
+        if (!($twoFactorToken instanceof TwoFactorTokenInterface)) {
             return;
         }
 
-        $providerName = $this->twoFactorToken->getCurrentTwoFactorProvider();
+        $providerName = $twoFactorToken->getCurrentTwoFactorProvider();
         if (null === $providerName) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


mathieudz:
Maybe it would be even better to make the listener stateless. It could fetch the token from TokenStorage instead.

scheb:
Thanks for addressing that. Making the class fully stateless wouldn't unfortunately work, as you'd still need to have some flag variable to be set by the event callbacks. The way you implemented it by unsetting the token variable does the job imo.

I'd ask you for two changes:

Please move the resetting of the token after the first if statement, to make sure the token is always reset by the master request.
https://github.com/scheb/2fa/blob/e557f988ab1104fa9f27a7624d3f5c67319cb7ae/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php#L106-L108
You'd need to store the token in a local variable to make it work.

Would you please update the test suite to test the token is actually been reset when the method is called by the master request.

Thanks and happy holidays!

mathieudz:
@scheb How do you want me to check that the token has been reset? Add a public accessor just for that test?

scheb:
No, I wouldn't go for an artificial public accessor. You could rather test its behavior. Trigger `onKernelResponse` twice, `PreparationRecorderInterface` should be called once though.

Unfortunately Travis-CI stopped working yesterday, therefore the build isn't working currently. Please bear with me, I'm about to migrate CI to GitHub Actions now. I'll ping you as soon as this is done, so you can rebase your chances to get a CI run.

mathieudz:
I have extended an existing test. Not sure if you rather have it as a separate test, but it would have mainly duplicate code.

scheb:
It's fine the way you did it. The test is testing the result of the "happy case", so that's part of it.

CI has been migrated to now, so please rebase your commits onto 5.x and force push to trigger a proper build.

mathieudz:
Updated as requested

scheb:
Merged and released as  v5.3.1.

Thanks for contribution and a happy new year!

mathieudz:
Thanks for releasing it straight away!

PR #44, opened 2020-12-26
Make it bold that custom configurations will not be compatible

PhilETaylor:
This screwed me over. Lets make it bold to prevent someone else having the same issue.


scheb:
Thanks!

PR #45, opened 2020-12-29
Added support for lcobucci/jwt v4

wouterj:
Some major things changed in v4 unfortunately. This Pr adds support for v4 on top of v3. As you can see, it adds quite some complexity. I'll let you decide whether this complexity is worth it - or if v3 support should be dropped, I can update the PR accordingly.

ea7c2ad7afb6aaed08dfe821341b0d3e0d39f856

### DIFF
--- a/src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
+++ b/src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace Scheb\TwoFactorBundle\Security\TwoFactor\Trusted;
 
-use Lcobucci\JWT\Builder;
-use Lcobucci\JWT\Parser;
+use Lcobucci\JWT\Configuration;
+use Lcobucci\JWT\Exception;
 use Lcobucci\JWT\Signer\Hmac\Sha256;
-use Lcobucci\JWT\Token;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use Lcobucci\JWT\Token\Plain;
+use Lcobucci\JWT\Validation\Constraint\SignedWith;
 
 class JwtTokenEncoder
 {
@@ -16,51 +18,45 @@ class JwtTokenEncoder
     public const CLAIM_VERSION = 'vsn';
 
     /**
-     * @var Sha256
+     * @var Configuration
      */
-    private $signer;
-
-    /**
-     * @var string
-     */
-    private $applicationSecret;
+    private $configuration;
 
     public function __construct(string $applicationSecret)
     {
-        $this->signer = new Sha256();
-        $this->applicationSecret = $applicationSecret;
+        $this->configuration = Configuration::forSymmetricSigner(new Sha256(), InMemory::plainText($applicationSecret));
+        $this->configuration->setValidationConstraints(new SignedWith($this->configuration->signer(), $this->configuration->signingKey()));
     }
 
-    public function generateToken(string $username, string $firewallName, int $version, \DateTimeInterface $validUntil): Token
+    public function generateToken(string $username, string $firewallName, int $version, \DateTimeImmutable $validUntil): Plain
     {
-        $builder = (new Builder())
-            ->setIssuedAt(time())
-            ->setExpiration($validUntil->getTimestamp())
-            ->set(self::CLAIM_USERNAME, $username)
-            ->set(self::CLAIM_FIREWALL, $firewallName)
-            ->set(self::CLAIM_VERSION, $version)
-            ->sign($this->signer, $this->applicationSecret);
+        $builder = $this->configuration->builder()
+            ->issuedAt(new \DateTimeImmutable())
+            ->expiresAt($validUntil)
+            ->withClaim(self::CLAIM_USERNAME, $username)
+            ->withClaim(self::CLAIM_FIREWALL, $firewallName)
+            ->withClaim(self::CLAIM_VERSION, $version);
 
-        return $builder->getToken();
+        return $builder->getToken($this->configuration->signer(), $this->configuration->signingKey());
     }
 
-    public function decodeToken(string $token): ?Token
+    public function decodeToken(string $token): ?Plain
     {
         try {
-            $token = (new Parser())->parse($token);
-        } catch (\InvalidArgumentException $e) {
+            $token = $this->configuration->parser()->parse($token);
+        } catch (Exception $e) {
             return null; // Could not decode token
         }
 
-        try {
-            if (!$token->verify($this->signer, $this->applicationSecret)) {
-                return null;
-            }
-        } catch (\BadMethodCallException $e) {
+        if (!$token instanceof Plain) {
             return null;
         }
 
-        if ($token->isExpired()) {
+        if (!$this->configuration->validator()->validate($token, ...$this->configuration->validationConstraints())) {
+            return null;
+        }
+
+        if ($token->isExpired(new \DateTimeImmutable())) {
             return null;
         }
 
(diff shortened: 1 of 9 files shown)


scheb:
Haha, great, I also had that on my list work on that during the holidays. Thanks for looking into it!

I've seen you put quite some effort in making it compatible with both version 3 and 4 of lcobucci/jwt. I'd actually have no problem with bumping the dependency requirement up to `^3.4|^4.0` so you implement against the new interface (3.4 has a compatibility layer to 4 as far as I've seen). Should simplify things a lot. At least that's how I would have done it.

src/trusted-device/Security/TwoFactor/Trusted/TrustedDeviceToken.php
```diff
@@ -4,38 +4,38 @@
...
 class TrustedDeviceToken
 {
     /**
-     * @var Token
+     * @var Plain
      */
     private $jwtToken;
 
-    public function __construct(Token $jwtToken)
+    public function __construct(Plain $jwtToken)
```
wouterj, line 16: Officially, this class is not internal, do you consider this a BC break? If you do, we can leave the typehint as-is (v4 has `Token` as interface) and throw an `InvalidArgumentException` if it's not a plain token (we require a plain token, as `claims()` is not in the `Token` interface)

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
```diff
@@ -16,51 +19,56 @@ class JwtTokenEncoder
...
-            $token = (new Parser())->parse($token);
-        } catch (\InvalidArgumentException $e) {
+            $token = $this->configuration->parser()->parse($token);
+        } catch (Exception $e) {
             return null; // Could not decode token
         }
 
+        if (!$token instanceof Plain) {
+            return null;
+        }
```
wouterj, line 53: I've added the check for `Plain` in this class. Although it's not strictly necessary, all code using `decodeToken()` requires it to be a plain token. Do you also want me to change the return type to `?Plain`? (this would remove the exception in `src/trusted-device/Security/TwoFactor/Trusted/TrustedDeviceTokenEncoder.php:32`)

wouterj:
> I've seen you put quite some effort in making it compatible with both version 3 and 4 of lcobucci/jwt. I'd actually have no problem with bumping the dependency requirement up to `^3.4|^4.0` so you implement against the new interface (3.4 has a compatibility layer to 4 as far as I've seen). Should simplify things a lot. At least that's how I would have done it.

Oh, good idea! I've updated the PR accordingly.

Don't worry about my time. I actually took a shortcut by just copying the original class and rewritting things according to the nice diff examples in the upgrade guide (without looking if there was a nicer way). I didn't even know 3.4 had this nice BC layer.

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
```diff
@@ -16,51 +19,55 @@ class JwtTokenEncoder
...
+
         try {
-            if (!$token->verify($this->signer, $this->applicationSecret)) {
+            if (!$this->configuration->validator()->validate($token, ...$this->configuration->validationConstraints())) {
                 return null;
             }
         } catch (\BadMethodCallException $e) {
```
scheb: I believe this check can go away, the new implementation doesn't seem to throw that exception anymore.

tests/Security/TwoFactor/Trusted/JwtTokenEncoderTest.php
```diff
@@ -4,41 +4,47 @@
...
+use Lcobucci\JWT\Configuration;
 use Lcobucci\JWT\Signer\Hmac\Sha256;
-use Lcobucci\JWT\Token;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use Lcobucci\JWT\Token\Plain;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Trusted\JwtTokenEncoder;
 use Scheb\TwoFactorBundle\Tests\TestCase;
 
-class JwtTokenEncoderTest extends TestCase
+class AbstractJwtTokenEncoderTest extends TestCase
```
scheb: I believe this is a leftover from the previous implementation, right? Could be renamed back to `JwtTokenEncoderTest`.

tests/Security/TwoFactor/Trusted/JwtTokenEncoderTest.php
```diff
@@ -4,41 +4,47 @@
...
 {
-    private const CLAIM = 'test';
-    private const TOKEN_ID = 'tokenId';
-    private const APPLICATION_SECRET = 'applicationSecret';
+    protected const CLAIM = 'test';
+    protected const TOKEN_ID = 'tokenId';
+    protected const APPLICATION_SECRET = 'applicationSecret';
```
scheb: And these could be changed back to `private`.

tests/Security/TwoFactor/Trusted/TrustedDeviceTokenTest.php
```diff
@@ -18,19 +21,31 @@ class TrustedDeviceTokenTest extends TestCase
...
+                ->willReturnMap([
+                    [JwtTokenEncoder::CLAIM_USERNAME, false, 'username'],
+                    [JwtTokenEncoder::CLAIM_FIREWALL, false, 'firewallName'],
+                    [JwtTokenEncoder::CLAIM_VERSION, false, 1],
+                ]);
+            $jwtToken
+                ->expects($this->any())
+                ->method('__toString')
+                ->willReturn('encodedHeaders.encodedClaims.');
+        }
```
scheb: Also looks like a leftover from the previous implementation. Could be removed, right?

src/trusted-device/Security/TwoFactor/Trusted/TrustedDeviceToken.php
scheb: I consider these classes (my classes) internal classes, they're an implementation detail rather than an extension point. So I'm fine with changing the method signature. Therefore let's keep `Plain` here.

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
scheb: As I've said in the other comment, I consider these classes implementation details rather than an extension points. So I'm fine with changing all signatures from `Token` to the `Plain` class.

scheb:
Even when you don't consider it much of an effort, thanks for doing it, much appreciated :)

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
wouterj: Alright, that's what I thought as well :)

I would recommend you to flag them with `@internal` or `@final` then. That communicates your considerations with these classes to everyone using your libraries (I've always been amazed by which classes of OSS projects people extend/override).

On top of this, the Symfony DebugClassLoader will also warn when debug=true people when they extend from an `@final` class or use an `@internal` class.

wouterj:
Updated once again. I've also changed the signature of `JwtTokenEncoder` to use `DateTimeImmutable` instead of `DateTimeInterval`, that removes the strange if..elseif...else as lcobucci/jwt only supports the immutable variants.

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
scheb: Yea, that same thought also came to my mind when I read your comment. You know know what crazy stuff people do with software. I just wasn't sure how to do it properly, so thanks for explaining.

scheb (approved):

scheb:
Merged, thank you!

Will go though my classes tomorrow to do the internal/final annotations before I publish a release.

Issue #54, opened 2021-02-06
Enable 2 factor module configuration via environment variables


aturki:
Right now configuring modules via environment variables is not supported for instance, the following configuration is does not work:

```yaml
scheb_two_factor:
  email:
    enabled: '%env(ENABLE_2FA)%'
```

Spomky:
What about the `bool` env processor ?
https://symfony.com/doc/current/configuration/env_var_processors.html#built-in-environment-variable-processors

```yaml
scheb_two_factor:
  email:
    enabled: '%env(bool:ENABLE_2FA)%'
```

aturki:
That was my first approach indeed. But it seems that Symfony is delaying the evaluation of these variables until later in the container's construction process. 
Since we are actually executing code in the bundle's configuration it's not possible to use the processors [ref](https://symfony.com/doc/current/configuration/using_parameters_in_dic.html)

scheb:
Ok, I just learned something new. Environment variables are only replaced after the container building phase and are represented by their placeholder string until then (which is in our case always evaluating to `true`).

When you need the environment variable value during container building, you have to take care of transforming the variables. `resolveEnvPlaceholders` (as suggested in your PR) seems to be the way to go: https://github.com/symfony/symfony/issues/21420#issuecomment-275459297

scheb:
Feature released with version 5.5.0.

PR #55, opened 2021-02-06
Enable modules configuration via env variables

aturki:
Implements support for modules configuration via env variables 
https://github.com/scheb/2fa/issues/54


src/bundle/DependencyInjection/SchebTwoFactorExtension.php
```diff
@@ -36,16 +36,31 @@ public function load(array $configs, ContainerBuilder $container): void
         }
 
         // Load two-factor modules
-        if (isset($config['email']['enabled']) && true === $config['email']['enabled']) {
-            $this->configureEmailAuthenticationProvider($container, $config);
+        if (isset($config['email']['enabled'])) {
+            $config['email']['enabled'] = (bool) $container->resolveEnvPlaceholders($config['email']['enabled'], true);
+
+            if (true === $config['email']['enabled']) {
```
scheb: I'd suggest to change this to

```php
if (isset($config['email']['enabled']) && $this->resolveFeatureFlag($config['email']['enabled'])) {
```

and add a `resolveFeatureFlag` method for the env-transformation and boolean casting logic. The name `resolveFeatureFlag` can be discussed, was just the first thing that came to my mind.

scheb (changes requested):

scheb:
Besides the changes requested above:

- Coding standards check is failing. Executing php-cs-fixer should solve this issue.
- Please add the same approach for `$config['trusted_device']['enabled']` and `$config['backup_codes']['enabled']` further down in the same file, as these are also boolean flags to enable bundle features.
- I'd like to have a test case to make sure environment variables are actually properly evaluated. This should cover both `true` and `false`-like values, to make sure env variable evaluation works as intended. For providing environment variables to test I'd suggest setting them via the `phpunit.xml.dist` configuration.

aturki:
@scheb Actually testing is not that straightforward. Since the service container is not build during unit tests, env variables are not evaluated correctly. Any pointers on this ?

src/bundle/DependencyInjection/SchebTwoFactorExtension.php
```diff
@@ -36,26 +36,30 @@ public function load(array $configs, ContainerBuilder $container): void
         }
 
         // Load two-factor modules
-        if (isset($config['email']['enabled']) && true === $config['email']['enabled']) {
+        if (isset($config['email']['enabled']) && true === $this->resolveFeatureFlag($container, $config['email']['enabled'])) {
```
scheb: The `true === ` can be removed. `resolveFeatureFlag` is only returning booleans, so checking for `true` is redundant here.

scheb (changes requested):

scheb:
> @scheb Actually testing is not that straightforward. Since the service container is not build during unit tests, env variables are not evaluated correctly. Any pointers on this ?

I was thinking of cases built similar to `load_emailAuthEnabled_loadEmailServices`. You overwrite `$config['email']['enabled']` to reference an environment variable (which can be predefined via phpunit.xml.dist`). For the `true` case it would define these services, for the `false` case it wouldn't. That way we can ensure the environment variable was properly evaluated.

aturki:
> I was thinking of cases built similar to `load_emailAuthEnabled_loadEmailServices`. You overwrite `$config['email']['enabled']` to reference an environment variable (which can be predefined via phpunit.xml.dist`). For the `true`case it would define these services, for the`false` case it wouldn't. That way we can ensure the environment variable was properly evaluated.

I pushed the test to better explain the problem. 
Basically it fails because the bundles configuration expects a boolean value for the `enabled` fields but without compiling the container, an expression like `%env(ENABLE_2FA)%` is not interpreted and considered a string value.

scheb:
Ok, I just tried myself with your code. Indeed a `%env(ENABLE_EMAIL_2FA)%` isn't accepted by the configuration because it's considered a "string" type, but the configuration node is expecting a boolean. Also, an actual Symfony application fails exactly the same, so it's not a problem with the test setup.

A workaround to this problem seems to be hinting the environment variable as a boolean `%env(bool:ENABLE_EMAIL_2FA)%`, which is accepted by the configuration, but then I get the following exception:

```
 Using a cast in "%env(bool:ENABLE_EMAIL_2FA)%" is incompatible with resolution at compile time in "Scheb\TwoFactorBundle\DependencyInjection\SchebTwoFactorExtension". The logic in the extension should be moved to a compiler pass,  
   or an env parameter with no cast should be used instead.
```

What a bummer. I don't really get why it's not possible to use type hinted environment variable at this stage, but well, that's how things are obviously are. The solution to this issue is of course to not use type hinting, and so we're back at the beginning XD

So how to fix this? I can think of 2 solutions right now:

1) Refactor the whole logic for optional features into compiler passes, as suggested by the exception message.
2) Or make the `enabled` configuration nodes accept non-boolean values (make them scalar nodes). Since they're now cased to boolean in `SchebTwoFactorExtension.php` things would still work the same but you'd be able to pass in the environment variable placeholder strings.

I'm not really excited to refactor all that code into compiler passes just to make environment variables work. The other solution feels a bit hacky but would be more pragmatic.

tests/DependencyInjection/SchebTwoFactorExtensionTest.php
```diff
@@ -107,6 +107,30 @@ public function load_fullConfig_setConfigValues(): void
         $this->assertHasParameter(['127.0.0.1'], 'scheb_two_factor.ip_whitelist');
     }
 
+    /**
+     * @test
+     */
+    public function load_envVarBasedFullConfig_setConfigValues(): void
```
scheb: This should be two test methods, one for the `true` case, one for the `false` case.

tests/DependencyInjection/SchebTwoFactorExtensionTest.php
```diff
@@ -491,6 +515,109 @@ private function getFullConfig(): array
         return $parser->parse($yaml);
     }
 
+    private function getEnvVarBasedFullConfig(bool $truthyConfig): array
+    {
+        if (true === $truthyConfig) {
+            $yaml = <<<EOF
```
scheb, line 527: You don't need a full configuration for the test. Just setting the `enabled` nodes is good enough.

tests/DependencyInjection/SchebTwoFactorExtensionTest.php
```diff
@@ -107,6 +107,30 @@ public function load_fullConfig_setConfigValues(): void
...
+    public function load_envVarBasedFullConfig_setConfigValues(): void
+    {
+        $config = $this->getEnvVarBasedFullConfig(true);
+        $this->extension->load([$config], $this->container);
+
+        $this->assertHasParameter(true, 'scheb_two_factor.trusted_device.enabled');
+        $this->assertHasParameter(true, 'scheb_two_factor.backup_codes.enabled');
+        $this->assertHasParameter(true, 'scheb_two_factor.google.enabled');
+        $this->assertHasParameter(true, 'scheb_two_factor.totp.enabled');
+        $this->assertHasParameter(true, 'scheb_two_factor.email.enabled');
```
scheb: Except for `scheb_two_factor.trusted_device.enabled`, these parameters don't exist. Instead, check for service definitions which should be set when the feature is enabled. For example, when email authentication is enabled, a service definition `scheb_two_factor.security.email.provider` should be defined. If it's disabled, this one isn't defined. For each of the feature toggles there is such services that you can use to check if the feature has been activated or not. Check the test methods for examples. One service defintion per feature flag is enough, no need to test all of them for each feature flag.

scheb (changes requested):

scheb:
Thanks for the update. I want to double check it's gonna be fine to do that config change. Hope to find some time later this week.

scheb:
Looks good.

I was playing around with the config, one final change request: Let's keep the `->canBeEnabled()` everywhere. It is adding some extra features that I'd like to keep. We should just overwrite the `enabled` node with a scalar node.

scheb (approved):

scheb:
Merged, thank you for this feature! Just released as version 5.5.0.

aturki:
Glad to be of help ;)

PR #68, opened 2021-03-31
Remove deprecation notice about getProviderKey in Symfony 5.2+

Seldaek:
Fixes thie deprecation notice:

> User Deprecated: Since symfony/security-http 5.2: Method "Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorToken::getProviderKey()" has been deprecated, rename it to "getFirewallName()" instead.


scheb:
Awesome, thanks!

Will look into your other two issues over the long weekend.

And btw, thank you for all your work over the last years for the PHP community! Much appreciated! :)

PR #72, opened 2021-04-15
Altered backUpCode type

salirzv:
**Description**

The json_array type is deprecated and it should have a default value 
I also added a setter method to add new back up codes
The default value of backup_codes in config file should be true, because user is here to enable the bundle (IMO)


scheb:
Thanks for keeping that up-to-date!

Issue #67, opened 2021-03-31
Badge support for AuthenticationManager and API 2fa-auth


Seldaek:
Context: I am currently migrating a project to the experimental AuthenticationManager as well as using this bundle instead of a homegrown 2fa solution.

We do have an API with an existing endpoint for logging in users which returns a special code to the client if 2fa is required, and then expects the login API call again with username, password, AND the 2fa code all present in one to do authentication in a single request.

This does not really play well with your API guidelines which require a two-request flow, which IMO is more complex as the client needs to keep even more state.

Anyway I ended up implementing this myself as it isn't that hard, but I also got pretty familiar with the new auth code as I had to do various authenticator implementations. So I am mostly posting this here for others who may need the help to figure this out.. and I am thinking if this bundle offered better support for it in the future it would maybe be good (could be a v6 thing for sure).

I now create this badge in my API authenticator, which gets added on the passport:

```
new TwoFactorAuthBadge($credentials['2fa_code'] ?? null)
```

The badge class is a very simple value object:
```php
<?php

class TwoFactorAuthBadge implements BadgeInterface
{
    private $resolved = false;
    private ?string $twoFaCode;

    public function __construct(?string $twoFaCode)
    {
        $this->twoFaCode = $twoFaCode;
    }

    public function getTwoFaCode(): ?string
    {
        return $this->twoFaCode;
    }

    /**
     * @internal
     */
    public function markResolved(): void
    {
        $this->resolved = true;
    }

    public function isResolved(): bool
    {
        return $this->resolved;
    }
}
```

And the auth listener which checks the badge is also not that complex, which is why I think it might be worth including here:

```php
<?php

use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Totp\TotpAuthenticatorInterface;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\Security\Http\Event\CheckPassportEvent;
use Symfony\Component\Security\Http\Authenticator\Passport\Passport;

class TwoFactorAuthListener implements EventSubscriberInterface
{
    private TotpAuthenticatorInterface $totpAuthenticator;

    public function __construct(TotpAuthenticatorInterface $totpAuthenticator)
    {
        $this->totpAuthenticator = $totpAuthenticator;
    }

    public function checkPassport(CheckPassportEvent $event): void
    {
        /** @var Passport $passport */
        $passport = $event->getPassport();
[...]

PR #74, opened 2021-04-16
Document how to bypass 2FA entirely in a given context

Seldaek:
This is partially a question too.. is this the best way? 

I first tried setting up a custom 2fa condition class but I did not have access to the Passport via the AuthenticationContextInterface, so it made things tricky as I couldn't easily add a TwoFactorBypassBadge or something to look at. I had to use the token, then when looking at internals further I found out that I may not even need a custom condition at all if I'm modifying the token already.. But IMO a passport badge would be more natural in the new authenticator.

### DIFF
--- a/doc/custom_conditions.md
+++ b/doc/custom_conditions.md
@@ -27,3 +27,27 @@ Register it as a service and configure the service name:
 scheb_two_factor:
     two_factor_condition: acme.custom_two_factor_condition
 ```
+
+Bypassing Two-Factor Authentication
+===================================
+
+If you simply wish to bypass 2FA in a given Authenticator context, setting the 
+`TwoFactorAuthenticator::FLAG_2FA_COMPLETE` attribute on the token will achieve this.
+
+For example, if you are building a [custom Authenticator](https://symfony.com/doc/5.2/security/experimental_authenticators.html#creating-a-custom-authenticator)
+this would bypass 2FA when the authenticator is used:
+
+```php
+class MyAuthenticator extends AbstractAuthenticator
+{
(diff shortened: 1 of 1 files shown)


scheb:
So you can't make the decision whether 2fa should be applied or nor based on the security token, but you'd want to make it based on the authenticator that was used to generate that security token? I guess you have multiple authenticators all using the same security token class. Do I understand that right?

Seldaek:
Yes that is correct. I could for sure also create a new token, but IMO the token should be used to represent a user and shouldn't have to change here. The passport/badge system is used to configure the auth flow and so it would be much more natural to do it at that level. This could be a feature request I guess, but I needed a solution now, and this attribute on the token appears to work :)

scheb:
Attribute on token is in fact a valid solution imo.

I'd need to look into what the flow with the new authenticator system is like. The decision is based on the token mostly because that was the only way to distinguish authentication methods in the classic security system.

Seldaek:
Yup works fine for current setup. In a future world where there is only authenticator. I think having a badge makes more sense, actually this is kinda covered by https://github.com/scheb/2fa/issues/67 already - if the TwoFactorAuthBadge was part of this bundle and recognized by it, then I could have added the badge, called markResolved() on it already in the authenticator to make it skip the badge auth step.

scheb:
I'll add a note that this approach is only working with the authenticator system :)

If the PR for Symfony is accepted I can include the passport in the `AuthenticationContext`. Then you have a badge on the passport and check for it in a custom condition. In my opinion that would be the clean solution.

Issue #86, opened 2021-05-22
Allow environment variable use in configuration


gnat42:
## Problem Desc
We have 2fa enabled on our production instance. We use the IP whitelist so that local development disables 2FA. Recently we've started using a centralized QA server. This server doesn't send email or any outbound messages. It uses the symfony prod environment. So I'd like to disable the email via an environment parameter. I created one that is in our .env file PROJECT_ENABLE_2FA=false and edited 

``` 
//config/packages/scheb_two_factor.yaml
scheb_two_factor:
    email:
        enabled: '%env(PROJECT_ENABLE_2FA)%'
```
However when I do this email auth is still attempting to send a 2fa and I'm presented with a request for the code. Using xdebug I could see that the code in [SchebTwoFactorExtenstion.php](https://github.com/scheb/2fa/blob/5.x/src/bundle/DependencyInjection/SchebTwoFactorExtension.php#L39) should be disabling it, however the code
```
$this->resolveFeatureFlag($container, $config['email']['enabled'])
```

returns the string "false" so it's not actually false because if("string") isn't a falsey value. I looked at some docs and then tried

``` 
//config/packages/scheb_two_factor.yaml
scheb_two_factor:
    email:
        enabled: '%env(bool:PROJECT_ENABLE_2FA)%'
```

However now I get an error

```
Using a cast in "env(bool:PROJECT_ENABLE_2FA)" is incompatible with resolution at compile time in "Scheb\TwoFactorBundle\DependencyInjection\SchebTwoFactorExtension". The logic in the extension should be moved to a compiler pass, or an env parameter with no cast should be used instead.
````

I don't know the right way to solve this. I could submit a pull request so that this->resolveFeatureFlag if not a boolean then looks for the string "true"|"false"|"on"|"off" and returns an appropriate boolean otherwise it returns the value. I think that should be backwards compatible in the library. I just don't know if the symfony error its throwing points to a better way to solve it. I'm not a huge fan of the environment variable addition to symfony as I keep running into weird issues like this that didn't occur with parameters.yml files. I know that's not your fault, I'm just point it out as I don't know the correct solution. If you're happy enough with a PR to fix this here I will happy draft one up for you quickly.

gnat42:
Ok, I'm thoroughly confused now. I wrote up a PR where I changed that function, in submitting it I saw the request for a test so I looked at existing tests and you are testing for env variables!!! So that was really bizarre. Next thing I looked at is this:

https://3v4l.org/B8Qel

[...]

PR #87, opened 2021-05-22
Handle string env vars

gnat42:
**Description**

When you use environment variables in a configuration, they are resolved by default as strings. It isn't obvious that setting an env var to false creates a string and so its hard to know why it doesn't work. false/true are logically used defaults. Symfony doesn't allow us to use env(bool:VAR). 

Therefore we check for the string variables "false"|"off"|"true"|"on" and if found return the appropriate boolean. 

Fixes issue #86

### DIFF
--- a/tests/DependencyInjection/SchebTwoFactorExtensionTest.php
+++ b/tests/DependencyInjection/SchebTwoFactorExtensionTest.php
@@ -140,6 +140,60 @@ class SchebTwoFactorExtensionTest extends TestCase
     /**
      * @test
      */
+    public function load_offOrFalseStringEnvVarBasedConfig_setConfigValues(): void
+    {
+        $yaml = <<<EOF
+trusted_device:
+    enabled: "%env(ENABLE_2FA_OFF_STR)%"
+backup_codes:
+    enabled: "%env(ENABLE_2FA_OFF_STR)%"
+email:
+    enabled: "%env(ENABLE_2FA_OFF_STR)%"
+google:
+    enabled: "%env(ENABLE_2FA_OFF_STR)%"
+totp:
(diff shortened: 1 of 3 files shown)


tests/DependencyInjection/SchebTwoFactorExtensionTest.php
```diff
@@ -137,6 +137,60 @@ public function load_falsyEnvVarBasedConfig_setConfigValues(): void
...
+     */
+    public function load_onOrTrueStringEnvVarBasedConfig_setConfigValues(): void
+    {
+        $yaml = <<<EOF
+trusted_device:
+    enabled: "%env(ENABLE_2FA_ON_STR)%"
+backup_codes:
+    enabled: "%env(ENABLE_2FA_ON_STR)%"
+email:
+    enabled: "%env(ENABLE_2FA_OFF_STR)%"
```
scheb: Shouldn't that be on as well?

gnat42: I wanted to test a 'non-uniform' set of enable/disable settings, but it doesn't really have to be. Let me know if you'd prefer it all on/off and I'll make the change

scheb: Yea, please change that. The falsy case has already been covered by the test above, so we'd miss to cover the truthy case :)

Besides that 👍

gnat42: makes sense and sounds good. I've made the change and pushed the branch again

scheb:
Thanks, I'll fix the Psalm errors and then publish a release.

scheb:
Released as 5.9.0

PR #88, opened 2021-05-24
Fix deprecated in Symfony 5.3

ben29:
Fixing:

> User Deprecated: Since symfony/http-kernel 5.3: "Symfony\Component\HttpKernel\Event\KernelEvent::isMasterRequest()" is deprecated, use "isMainRequest()" instead.

### DIFF
--- a/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
+++ b/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
@@ -110,7 +110,7 @@ class TwoFactorProviderPreparationListener implements EventSubscriberInterface
             return;
         }
 
-        if (!$event->isMainRequest()) {
+        if (!$event->isMasterRequest()) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


scheb:
Hey there, thanks for the PR, but as you can see from the failing tests, you can't just exchange the method with the newer one. The bundle is supporting Symfony versions down until 4.4, so the solution has to support all these older versions as well.

So usually you'd check with `method_exists(KernelEvent::class, 'isMainRequest')` if that new method is available, then use it. Otherwise use the old method.

And please don't forget to update the test cases.

src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
```diff
@@ -106,7 +106,11 @@ public function onTwoFactorForm(TwoFactorAuthenticationEvent $event): void
 
     public function onKernelResponse(ResponseEvent $event): void
     {
-        if (!$event->isMasterRequest()) {
+        if (method_exists(KernelEvents::class, 'isMainRequest') && !$event->isMainRequest()) {
+            return;
+        }
+
+        if (!$event->isMainRequest()) {
```
scheb: Now you're using only `isMainRequest`, but you need to fall back to `isMasterRequest` when `isMainRequest` isn't available.

ben29: sorry my mistake!

PR #89, opened 2021-05-24
Update TwoFactorProviderPreparationListener.php

ben29:
fixing:

> User Deprecated: Since symfony/http-kernel 5.3: "Symfony\Component\HttpKernel\Event\KernelEvent::isMasterRequest()" is deprecated, use "isMainRequest()" instead.

on Symfony 5.3

PR #90, opened 2021-06-02
check for isMainRequest on the correct class

PhilETaylor:
As commented in https://github.com/scheb/2fa/pull/89#issuecomment-847347811

> While tests are green now, this isn't fixing the deprecation issue. The method isMasterRequest is still called when isMainRequest is true.

Looks like you made a simple typo and reused the already imported plural class name `KernelEvents`

Today in my Symfony 5.3.0 project I have this logged deprecation in my dev env.

Screenshot 2021-06-02 at 11 27 18

On checking `isMainRequest` is not a method in `Symfony\Component\HttpKernel\KernelEvents` (the plural of events) but is a method on `Symfony\Component\HttpKernel\KernelEvent` (the singular of event)

This PR corrects the typo, which in turn allows the correct evaluation in `method_exists` and then correctly calls `isMainRequest ` and therefore no more logged deprecation.

### DIFF
--- a/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
+++ b/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php
@@ -11,7 +11,7 @@ use Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorAuthenticationEvent;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorAuthenticationEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
-use Symfony\Component\HttpKernel\KernelEvent;
+use Symfony\Component\HttpKernel\Event\KernelEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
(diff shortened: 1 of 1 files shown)


PhilETaylor:
I dont believe the tests failing is the fault of this PR?

scheb:
Oh damn it, thanks for fixing this. The build errors are not related to your changes. Symfony people re-introduced the `getUsername` method with Symfony 5.3.1 and the checks don't like that.

PhilETaylor:
No worries - now running Symfony 5.3.1 and scheb\2fa in production - fingers still crossed ;-)

scheb:
Yea, neither issue makes any difference for you in prod.

It's mocking framework and code analysis complaining about something that shouldn't happen. Agree with Symfony people that it's better to reintroduce those methods. Found it odd that they just removed them, because to me that was clearly a BC break, but it was supposed to be "normal practise". Yea whatever, I can probably remove this beautiful piece of code again :D

https://github.com/scheb/2fa/blob/5.x/tests/TestCase.php#L30-L79

scheb:
Waiting for symfony/symfony#41521 to be solved with Symfony 5.3.2, so I can run a proper build. Fortunately we're not in a hurry with this fix :)

PhilETaylor:
>  Fortunately we're not in a hurry with this fix :)

The isMainRequest fix was just to get rid of the deprecated notice in my logs - totally no rush for that... as for the tests running... well, we all need a bit of fire to keep warm :)

scheb:
Released as v5.10.1

PhilETaylor:
Thanks

PR #98, opened 2021-07-02
Return roles of authenticatedToken in TwoFactorToken

Zombaya:
**Description**
The `TwoFactorToken` currently does not return the roles of the `authenticatedToken`.  This resulted in a bug in our application.  To resolve this, I modified the class to pass through the roles of the underlying token.

***Our problematic situation***
1. We have a `AuthenticationSuccessSubscriber` listening to `AuthenticationEvents::AUTHENTICATION_SUCCESS`
2. In this subscriber, we check if the logged in user has access to certain roles using `\Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface::isGranted()`
3. This propagates the decision to `Symfony\Component\Security\Core\Authorization\Voter\RoleVoter`
4. This voter fetches the roles from the supplied token, which was the `TwoFactorToken`
5. The TwoFactorToken returned an empty array, implying the current user does not have any roles available.
6. The RoleVoter denies access to the role

***With new implementation***
5. The TwoFactorToken returns the roles  of the authenticatedToken
6. The RoleVoter grants access to the role

***Additional details***
Previously, if the specific user did not have 2fa configured, it would work as expected, since then the supplied token would not be a TwoFactorToken and the RoleVoter had access to the roles of the user.

This resulted in a different result from the RoleVoter depending on the user having 2fa configured or not.

***Follow up***
If you have any more questions or need some implementation of this PR changed, I'm happy to help.

I added an entry to `.gitignore` since php-cs-fixer generated a different cache-file for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ vendor/
 composer.lock
 # PHPCS Fixer
 .php-cs-fixer.cache
+.php_cs.cache
 # PHPUnit
 .phpunit.result.cache
 # Test application
--- a/src/bundle/Security/Authentication/Token/TwoFactorToken.php
+++ b/src/bundle/Security/Authentication/Token/TwoFactorToken.php
@@ -75,13 +75,51 @@ class TwoFactorToken implements TwoFactorTokenInterface
     // Compatibility for Symfony < 5.0
     public function getRoles(): array
     {
+        if (method_exists($this->authenticatedToken, 'getRoles')) {
+            return $this->authenticatedToken->getRoles();
+        }
+
+        if (method_exists($this->authenticatedToken, 'getRoleNames')) {
+            return $this->authenticatedToken->getRoleNames();
+        }
+
         return [];
     }
 
     // Compatibility for Symfony >= 4.3
+    // Must return array of strings
     public function getRoleNames(): array
     {
-        return $this->getRoles();
+        if (method_exists($this->authenticatedToken, 'getRoleNames')) {
+            return $this->authenticatedToken->getRoleNames();
+        }
+
+        if (method_exists($this->authenticatedToken, 'getRoles')) {
+            // Try to convert to string, remove any roles who failed to convert to a string
+            return array_filter(
+                array_map(
+                    /** @param mixed $role */
+                    static function ($role) {
+                        /** @psalm-suppress UndefinedClass */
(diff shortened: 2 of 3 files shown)


Zombaya:
I'll take a look at the failed unit-tests.  I wasn't too sure about them to be honest.

scheb:
Short response, because I'm at work and don't have much time: This approach cannot work, it's breaking one of the essential concepts of the 2fa-bundle, which is holding back the roles until the user is fully authenticated. Can give you a more extensive answer later today.

Zombaya:
OK, thanks for the headsup.  If you can point me in the right direction on how I'd need to proceed or where I can find more information, I'm willing to invest more time into getting this working.

I've taken a quick glance at the TwoFactorToken again. Would it be an option to add a check to verify if the twoFactorAuthentication is fully completed?

```php
function getRoles: array 
{
     if (!$this->allTwoFactorProvidersAuthenticated()) {
        return [];
     }
     // rest of current implementation
}
```

scheb:
There's two ways how you can approach this problem:

1) You create your own variant of `TwoFactorTokenInterface`, which exposes the roles. Then you have to somehow deal with the problem that the user already has all the roles granted, even when they haven't completed 2fa yet. To do that implement your own variant of `TwoFactorTokenInterface` and `TwoFactorTokenFactoryInterface` and use the config option `two_factor_token_factory` to make the bundle use your own token factory service.

2) You ignore all events with a `TwoFactorToken` in your `AuthenticationSuccessSubscriber`, because at that point the user isn't fully authenticated yet, so no need to check for any roles. Later, when the user is completing 2fa, another success event will be dispatched and then you can execute that check.

scheb:
I should mention that I cannot accept this PR in any form. Holding back the roles is essential to the bundle's concept. There is no way that I can accept this into the bundle, because it would suddenly leak roles that shouldn't be granted and break 2fa for everyone who's currently using the bundle.

Zombaya:
I thank you for the thorough explanation on how we need to proceed.  I think I'll try tackling the problem using your second suggestion, as that looks to be the more correct implementation.

I completely understand that you are not able to accept this PR if it breaks the security-model of the bundle.  In my opinion this speaks to the quality of the bundle and I'd like to express my gratitude for the work you have put in and the result that you have created.

I found that the various checks and CI-tools guide you towards better code when creating this PR.  

If I'm able to come up with valuable documentation from the eventSubscriber we'll be creating, I'll send another PR about it.

In general: thanks!

PR #100, opened 2021-07-15
Avoid deprecation warnings in Symfony 5.3

Seldaek:
Not sure if method_exists is typically the way you'd do that, but hopefully this works.

### DIFF
--- a/src/bundle/Security/Authentication/Provider/AuthenticationProviderDecorator.php
+++ b/src/bundle/Security/Authentication/Provider/AuthenticationProviderDecorator.php
@@ -105,7 +105,12 @@ class AuthenticationProviderDecorator implements AuthenticationProviderInterface
 
     private function getRequest(): Request
     {
-        $request = $this->requestStack->getMasterRequest();
+        // Compatibility for Symfony >= 5.3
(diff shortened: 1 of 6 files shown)


scheb:
Obviously this hasn't been covered in #89 :/ There's obviously a whole bunch of these `isMasterRequest` still left: https://github.com/scheb/2fa/search?q=getMasterRequest

The `method_exists` approach is also what I'd use. A reference implementation is [here](https://github.com/scheb/2fa/blob/5.x/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderPreparationListener.php#L117-L126). Tests would need an update as well, since Symfony 5.3+ builds use the new code path.

If you could add `Compatibility for Symfony >= 5.3` as a comment would be wonderful, helps me to find these things easily once I'm able to clean it up in the next major version.

Seldaek:
Sure I'll try to get this done next week.

scheb:
No rush :)

I'd do it myself, but the new job is very demanding, currently no motivation to do coding after work :/

Seldaek:
Ok I think it's good to go :)

scheb:
Awesome, thank you Jordi!

PR #105, opened 2021-08-22
Minor reStructuredText syntax fixes/improvements

wouterj:
I quickly scanned all doc files and fixed/improved some reStructuredText syntax. A brief list of the main things:

- reST does not have a strict standard on what characters to use to mark title levels. In [the Symfony docs standards](https://symfony.com/doc/current/contributing/documentation/standards.html#sphinx) we recommend `~` for level 3, I think it reads nicer than `^` (but of course, you are free to not follow the docs standards)
- reST allows you to have "directive" (those warning/tip/caution/info blocks). I've added them wherever I believe that's what you actually want (e.g. you currently used the "i" character for things I think are `.. info::` or `.. tip:``)
- I'm not sure about the ordered list character, but we always use `#.` instead of `1), 2)` etc. in the Symfony docs.
- reST doesn't allow nesting in-line markup (e.g. bold italic or a literal inside a link).
- We support some custom highlighting languages (like `terminal` and `html+twig`). Also, if you show the same example in multiple formats, you can use a  ["configuration block"](https://symfony.com/doc/current/contributing/documentation/format.html#configuration-blocks) to create the tabs in your code example.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -28,13 +28,13 @@ least similar, as long as it lets you configure a custom success handler.
 
 You need to implement 4 classes:
 
-1) A custom success handler for the authentication mechanism
-2) A custom "two-factor authentication required" handler for the two-factor authentication
-3) A custom success handler for the two-factor authentication
-4) A custom failure handler for the two-factor authentication
+#. A custom success handler for the authentication mechanism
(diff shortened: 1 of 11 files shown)


scheb:
Thanks for the review and improvements! Happy to merge those in 🙇

Issue #25, opened 2020-09-14
Support hardware tokens (WebAuthn)


stephanvierkant:
I'm using hardware tokens from Yubico. It would be great if this library supports WebAuthn. 

I've used [r/u2f-two-factor-bundle](https://github.com/darookee/u2f-two-factor-bundle) before, but it doesn't support scheb/2fa and it lacks some quality this library has. And I've think it's better to support the newer WebAuthn API instead of FIDO2/U2F.

See https://caniuse.com/webauthn for browser support.

Spomky:
Hi @stephanvierkant,

Actually I started to develop an extension to that lib to support Webauthn.
It is based on the [web-auth/*](https://github.com/web-auth/webauthn-framework) framework I manage.
Unfortunately my planning has not allowed me to spend much time so far… but this is still in my todo list.

By the way, it will allow hardware tokens (Yubico, Feitian…) as well as internal authenticator (Android, Apple Face/TouchId…).

stephanvierkant:
Great to hear! Keep us posted and let us know if you need some help.

scheb:
I have to say I have no experience at all with hardware token (no idea what FIDO2/U2F is), so I'm probably not the best person to implement this.

Why not contribute back to `r/u2f-two-factor-bundle` to improve the library?

scheb:
Flagging this with "help wanted", if this should ever become part of the 2fa package and not being provided by a 3rd-party package. Someone with knowledge in that field, who knows how it's best implemented, what libraries to use, etc. would be needed to contribute it. As mentioned above, I'm feeling myself not experienced enough with that authentication method to implement it with confidence.

PhilETaylor:
> Flagging this with "help wanted"

I think its time I contributed back ;-) drop me a quick email at phil@phil-taylor.com and I'll reply back with my thoughts on a plan.

wouterj:
I have been playing with the `web-auth/*` packages and webauthn (using the Krypton app, as I don't have a physical key) before. 

From my findings, the best solution would be to create a two factor provider implementation using this package (or at least, webauthn is way to complex to implement without third party package). So I would suggest to either create a two factor provider for [`web-auth/webauthn-framework`](https://github.com/web-auth/webauthn-framework) in this bundle or someone contribute an implementation of `TwoFactorProviderInterface` to https://github.com/web-auth/webauthn-symfony-bundle

Spomky:
Hi @wouterj,

[...]

PR #106, opened 2021-08-30
Webauthn support

Spomky:
**Description**

This PR aims at providing a new 2FA based on Webauthn as per discussion in #25


scheb:
Thanks for sharing. I'll look into this at a later point. Weird that the integration tests are failing, don't really get why.

src/webauthn/Security/TwoFactor/Provider/Webauthn/WebauthnAuthenticatorTwoFactorProvider.php
```diff
@@ -0,0 +1,126 @@
...
+    public function validateAuthenticationCode($user, string $authenticationCode): bool
+    {
+        if (!($user instanceof WebauthnTwoFactorInterface)) {
+            return false;
+        }
+
+        $sfRequest = $this->getRequest();
+        $publicKeyCredentialRequestOptions = $sfRequest->getSession()->remove(self::WEBAUTHN_SESSION_PARAMETER_NAME);
+        if (null === $publicKeyCredentialRequestOptions) {
+            throw new \LogicException('No public key credential reauest options available');
```
scheb, line 87: Typo in `reauest` ;)

app/src/Repository/PublicKeyCredentialSourceRepository.php
```diff
@@ -0,0 +1,72 @@
...
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Persistence\ManagerRegistry;
+use Webauthn\PublicKeyCredentialSource as BasePublicKeyCredentialSource;
+use Webauthn\PublicKeyCredentialSourceRepository as PublicKeyCredentialSourceRepositoryInterface;
+use Webauthn\PublicKeyCredentialUserEntity;
+
+/**
+ * @ORM\Entity(repositoryClass="App\Repository\PublicKeyCredentialSourceRepository")
+ * @ORM\Table(name="public_key_credential_sources")
+ */
```
scheb, line 18: This can be removed from the repository class. Is this class actually needed? I don't see it used anywhere 🤔

app/src/Entity/User.php
```diff
@@ -78,6 +81,22 @@ class User implements UserInterface, \Serializable, EmailTwoFactorInterface, Goo
      */
     private $isActive;
 
+    /**
+     * @var Collection
+     * @ORM\OneToMany(targetEntity="PublicKeyCredentialSource", mappedBy="userHandle")
+     */
+    private $webauthnSources;
```
scheb, line 88: On login, this is causing an exception that it can't find `userHandle` on the other side.

Doctrine doesn't seem to take the additional fields inherited from `Webauthn\PublicKeyCredentialSource` into account. The schema just has the `id` field, nothing else :( I believe this is because the inherited fields don't have a `@ORM\Column` annotation, [though according to your documentation](https://webauthn-doc.spomky-labs.com/the-webauthn-server/the-symfony-way/entities-with-doctrine#doctrine-entity) it should work like that. Don't really know how to fix that.

Also, I believe, in order for the relation to work, the `userHandle` field on the other side would need to be an instance of the `User` type and not a string.

scheb (commented):
I finally found time to look into this. A few notes/questions.

What I'm not really getting is how this is wired up with Webauthn in the browser. Still needs some Javascript, right?

araab:
@Spomky Thank you for your work here and your other libs. Do you plan to finish this pull request?

scheb:
Closing due to inactivity. Maybe someone wants to pick this up in the future.

PR #109, opened 2021-10-06
TwoFactorFormListener: check if request has a session

brusch:
There could be situations where no session is given, in that case `$this->tokenStorage->getToken()` will throw an exception.
I think it would be better to just skip the handler in that case.

### DIFF
--- a/src/bundle/Security/TwoFactor/Event/TwoFactorFormListener.php
+++ b/src/bundle/Security/TwoFactor/Event/TwoFactorFormListener.php
@@ -44,6 +44,10 @@ class TwoFactorFormListener implements EventSubscriberInterface
 
     public function onKernelRequest(RequestEvent $requestEvent): void
     {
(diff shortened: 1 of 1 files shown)


scheb:
Oke, would you please add a test case for this scenario?

brusch:
@scheb there's no existing test for `TwoFactorFormListener`, still required?

scheb:
Ohhh, well, there should be one. I should definitely add that 🙈 So forget what I said, nothing for you, except you'd feel fancy to add it.

brusch:
@scheb alright, thanks - then I'll leave the PR as it is for now 😊

scheb:
Forgot to approve the build. So now it's throwing out some formatting issues. Would you please run php-cs-fixer? That should do it.

brusch:
@scheb fixed, but needs again approval 😊

PR #112, opened 2021-11-19
Fix Google Authenticator website URL

osavchenko:
**Description**

Update Google Authenticator website URL

### DIFF
--- a/doc/providers/google.rst
+++ b/doc/providers/google.rst
@@ -24,7 +24,7 @@ On successful authentication the bundle checks if there is a secret stored in th
 will ask for the authentication code. The user must enter the code currently shown in the Google Authenticator app to
 gain access.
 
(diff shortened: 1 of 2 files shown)


bytehead (approved):

scheb:
There's another occurrence of that URL in totp.rst, would you please update that one as well?

osavchenko:
Sure, done.
Also, I checked other URLs and found that one of the Symfony doc URLs got a redirect. I'm not sure if that redirect is ok or maybe there should be another URL. So, I changed the component version for now. Could you please advise how to fix this doc issue?

scheb:
The Symfony Docs URLs do not redirect for me. `current` would be what I'd want to keep. So please revert that.

scheb:
Thank you for contributing :)

PR #115, opened 2021-12-06
[Docs] Correct spelling in index.rst

gnito-org:


scheb:
Thanks for fixing that!

PR #116, opened 2021-12-06
[Docs] Change Google entity code example

gnito-org:
It's a change that PHPStorm Inspections recommended in my User entity.

### DIFF
--- a/doc/providers/google.rst
+++ b/doc/providers/google.rst
@@ -71,7 +71,7 @@ Authenticator for a user, generate a secret code and persist it with the user en
 
        public function isGoogleAuthenticatorEnabled(): bool
        {
-           return $this->googleAuthenticatorSecret ? true : false;
+           return (bool) $this->googleAuthenticatorSecret;
        }
 
        public function getGoogleAuthenticatorUsername(): string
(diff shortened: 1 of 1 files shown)


stephanvierkant:
And what if your `googleAuthenticatorSecret` is `00000`? Not sure if that is even possible, but (explicit) casting strings to boolean doesn't look that much better than implicit casting by using an if condition. See http://phpsadness.com/sad/47

What about `$this->googleAuthenticatorSecret !== null`?

scheb:
Changing it to a boolean cast wouldn't significantly change the logic. `null !== $this->googleAuthenticatorSecret` (Yoda Condition style) would actually be a change that I'd accept in :)

gnito-org:
Ok, I will modify the PR.

Just want to say that ``00000`` is not an issue because the secret is always a [string](https://github.com/scheb/2fa-google-authenticator/blob/fe0db75570b5f2de061e6c8446775ae90809a4e9/Security/TwoFactor/Provider/Google/GoogleAuthenticator.php#L44).

The current ``return $this->googleAuthenticatorSecret ? true : false;`` will also return false on a numeric ``00000``.

scheb:
Thanks for updating that so swiftly :)

PR #122, opened 2022-01-18
Fix deprecation warnings

Seldaek:
As Symfony now warns for return types that may be added in the future.. It'd be good to add these.

i.e. I got this in my test output:

```
  1x: Method "Symfony\Component\EventDispatcher\EventSubscriberInterface::getSubscribedEvents()" might add "array" as a native return type declaration in the future. Do the same in implementation "Scheb\TwoFactorBundle\Security\TwoFactor\Event\AuthenticationSuccessEventSuppressor" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "Symfony\Component\EventDispatcher\EventSubscriberInterface::getSubscribedEvents()" might add "array" as a native return type declaration in the future. Do the same in implementation "Scheb\TwoFactorBundle\Security\TwoFactor\Trusted\TrustedCookieResponseListener" now to avoid errors or add an explicit @return annotation to suppress this message.
```

There may be other places that need fixing, sorry if it's not complete.

### DIFF
--- a/src/bundle/Security/TwoFactor/Event/AuthenticationSuccessEventSuppressor.php
+++ b/src/bundle/Security/TwoFactor/Event/AuthenticationSuccessEventSuppressor.php
@@ -34,7 +34,7 @@ class AuthenticationSuccessEventSuppressor implements EventSubscriberInterface
     /**
      * {@inheritdoc}
      */
-    public static function getSubscribedEvents()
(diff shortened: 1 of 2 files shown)


scheb:
Oh I have missed those. Thanks! You definitely caught all the `getSubscribedEvents` that were missing the return type :)

PR #149, opened 2022-06-10
Docs: fix anonymous (public) role to define in firewall

jmsche:
Hi,

Little PR to fix the role to define on the firewall as it has changed, and the old one is not available anymore.

Changes should be okay as the bundle requires 5.4+/6.0+.


scheb:
Thanks for fixing that!

PR #153, opened 2022-07-12
Allow to use symfony mailer default configuration

alexander-schranz:
**Description**

The symfony/mailer allows already defining a default sender. Currently the scheb/2fa is not using it and use `no-reply@example.com` which can make issues if somebody is not changing it in there project. By defaulting to the `mailer` configuration it is less forgotten.

Symfony docs reference: https://symfony.com/doc/current/mailer.html#mailer-configure-email-globally

### DIFF
--- a/src/email/Mailer/SymfonyAuthCodeMailer.php
+++ b/src/email/Mailer/SymfonyAuthCodeMailer.php
@@ -14,16 +14,16 @@ use Symfony\Component\Mime\Email;
  */
 class SymfonyAuthCodeMailer implements AuthCodeMailerInterface
 {
-    private Address|string $senderAddress;
+    private Address|string|null $senderAddress = null;
 
     public function __construct(
         private MailerInterface $mailer,
-        string $senderEmail,
+        ?string $senderEmail,
         ?string $senderName,
     ) {
         if (null !== $senderName) {
             $this->senderAddress = new Address($senderEmail, $senderName);
-        } else {
+        } elseif ($senderEmail) {
             $this->senderAddress = $senderEmail;
         }
     }
@@ -38,9 +38,13 @@ class SymfonyAuthCodeMailer implements AuthCodeMailerInterface
         $message = new Email();
         $message
             ->to($user->getEmailAuthRecipient())
-            ->from($this->senderAddress)
             ->subject('Authentication Code')
             ->text($authCode);
+
+        if (null !== $this->senderAddress) {
+            $message->from($this->senderAddress);
+        }
+
         $this->mailer->send($message);
     }
 }
--- a/src/email/Mailer/SymfonyAuthCodeMailer.php
+++ b/src/email/Mailer/SymfonyAuthCodeMailer.php
@@ -21,7 +21,7 @@ class SymfonyAuthCodeMailer implements AuthCodeMailerInterface
         ?string $senderEmail,
         ?string $senderName,
     ) {
-        if (null !== $senderName) {
+        if (null !== $senderEmail && null !== $senderName) {
             $this->senderAddress = new Address($senderEmail, $senderName);
         } elseif ($senderEmail) {
             $this->senderAddress = $senderEmail;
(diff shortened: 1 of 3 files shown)


src/email/Mailer/SymfonyAuthCodeMailer.php
```diff
@@ -14,16 +14,16 @@
...
-    private Address|string $senderAddress;
+    private Address|string|null $senderAddress = null;
 
     public function __construct(
         private MailerInterface $mailer,
-        string $senderEmail,
+        ?string $senderEmail,
         ?string $senderName,
     ) {
         if (null !== $senderName) {
```
alexander-schranz:
```suggestion
        if (null !== $senderEmail && null !== $senderName) {
```

scheb:
Obviously some things changed in the ecosystem, causing all those failing builds (totally not your fault).

Please rebase on the latest 6.x branch to make those issues go away.

The last commit [Allow plugin for ci](https://github.com/scheb/2fa/pull/153/commits/2b17ac26dff2db9b31c22717fcc33b51861a1689) is actually not needed, once you've rebased, so please revert.

maxhelias (approved):

tests/Mailer/SymfonyAuthCodeMailerTest.php
```diff
@@ -59,6 +59,44 @@ public function sendAuthCode_hasAuthCode_sendEmail(): void
         $this->mailer->sendAuthCode($user);
     }
 
+    /**
+     * @test
+     */
+    public function sendAuthCode_hasAuthCode_sendEmail_withoutSender(): void
```
scheb: Please change this last part to `sendEmailWithoutSender`, thanks!

tests/Mailer/SymfonyAuthCodeMailerTest.php
```diff
@@ -59,6 +59,44 @@ public function sendAuthCode_hasAuthCode_sendEmail(): void
...
+        $user
+            ->expects($this->any())
+            ->method('getEmailAuthCode')
+            ->willReturn('1234');
+
+        $messageValidator = function ($mail) {
+            /** @var Email $mail */
+            $this->assertInstanceOf(Email::class, $mail);
+            $this->assertEquals('recipient@example.com', current($mail->getTo())->getAddress());
+            $this->assertNull(current($mail->getFrom()));
```
scheb: Seems to return `false` instead of `null`, `current()` of  empty array is `false`

scheb (changes requested):

tests/Mailer/SymfonyAuthCodeMailerTest.php
alexander-schranz: done :)

tests/Mailer/SymfonyAuthCodeMailerTest.php
alexander-schranz: done :)

scheb:
Looks like this change is not compatible with Symfony 5.4 versions :/

alexander-schranz:
@scheb for me it looks like more this is a problem of that this is a unit test which `mocks` the mailer. As the `5.4` docs represent that it should work also there: https://symfony.com/doc/5.4/mailer.html#email-addresses

alexander-schranz:
It just works with Symfony 6.1 because Symfony 6.1 supports draft emails: https://github.com/symfony/symfony/commit/0d6e859db236e37b8baa2b2bf4c8b7d14d151570 and validates not longer the sender and from. Still in 5.4 the sender is also set by: https://github.com/symfony/mailer/blob/v5.4.10/EventListener/EnvelopeListener.php#L46

Adopted the test to manually set the sender.

scheb:
Thanks for investigating this. I'm fine with that.

It's now phpcs being not happy with the positioning of that comment. I believe it you move it above the function declaration, it should work.

scheb:
Merged and released as  v6.1.0. Thanks for the effort!

alexander-schranz:
Thx for your help achieving this :)

Issue #157, opened 2022-08-21
lcobucci/jwt 4.2.0 b/c break - Key provided is shorter than 256 bits, only 240 bits provided


PhilETaylor:
**Bundle version**: 6.2.0
**Symfony version**: 6.1
**PHP version**: 8.1
**Using authenticators** (`enable_authenticator_manager: true`): YES

**Description**

lcobucci/jwt 4.2.0 has a b/c break (on purpose) which has resulted in exceptions

composer why lcobucci/jwt says scheb/2fa

`Lcobucci\JWT\Signer\InvalidKeyProvided: Key provided is shorter than 256 bits, only 240 bits provided`

Full stack trace https://sentry.io/share/issue/dc48c4847b5e4be68a51d69354cb5383/

**To Reproduce**

Its sporadic and DOESN'T happen on every login or use of 2fa. 

Its after login, but no idea yet anything other than the stack trace, we have been using scheb/2fa (and its previous versions) for many many years so some users might have very old setup of 2fa. 

It doesn't affect all users. Also if I set up a new 2fa on an account, I can login with no issues. Which is why I mention about "old setups". Also using 6 digit codes for full Google Authenticator compatibility in this app. 

Its on the post to /2fa_check with what appears to be a valid generated code supplied by the user

<img width="394" alt="Screen Shot 2022-08-21 at 12 21 38" src="https://user-images.githubusercontent.com/400092/185788568-7d7ee18c-5bae-4593-aa42-3b15c7e6c6da.png">

For now downgrading to `lcobucci/jwt: 4.1.5` seems to be enough to silence this - so that we can all enjoy our Sunday :) 

**Additional Context**

Release Notes https://github.com/lcobucci/jwt/releases/tag/4.2.0

Documentation on deprecation https://lcobucci-jwt.readthedocs.io/en/4.2.0/supported-algorithms/

lcobucci:
You may use the deprecated implementations that don't validate the key length (at least until you're able to rotate the keys)

PhilETaylor:
> You may use the deprecated implementations that don't validate the key length (at least until you're able to rotate the keys)

Yup - absolutely no rush here :-) 

> For now downgrading to lcobucci/jwt: 4.1.5 seems to be enough to silence this - so that we can all enjoy our Sunday :)

lcobucci:
[...]

PR #158, opened 2022-08-21
Configure JWT library via dependency injection

lcobucci:
**Description**

This addresses https://github.com/scheb/2fa/issues/157 by allowing users to configure the JWT generation process according to their needs. They can override the key, the algorithm, or the whole JWT configuration object without affecting the behaviour of the bundle.

### DIFF
--- a/app/config/services.yaml
+++ b/app/config/services.yaml
@@ -26,3 +26,9 @@ services:
 
     # add more service definitions when explicit configuration is needed
     # please note that last definitions always *replace* previous ones
+
+    scheb_two_factor.trusted_jwt_encoder.configuration.key:
+        class: Lcobucci\JWT\Signer\Key\InMemory
+        factory: [Lcobucci\JWT\Signer\Key\InMemory, 'base64Encoded']
+        arguments:
+            - oW2+MyKdgOf+iS1+qqgqb0FK5+1oZjDYGBemymEuzgU=
--- a/src/bundle/Resources/config/trusted_device.php
+++ b/src/bundle/Resources/config/trusted_device.php
@@ -2,6 +2,9 @@
 
 declare(strict_types=1);
 
+use Lcobucci\JWT\Configuration;
+use Lcobucci\JWT\Signer\Hmac\Sha256;
+use Lcobucci\JWT\Signer\Key\InMemory;
 use Scheb\TwoFactorBundle\Security\Http\EventListener\TrustedDeviceListener;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Condition\TrustedDeviceCondition;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Trusted\JwtTokenEncoder;
@@ -15,8 +18,21 @@ use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
-        ->set('scheb_two_factor.trusted_jwt_encoder', JwtTokenEncoder::class)
+        ->set('scheb_two_factor.trusted_jwt_encoder.configuration.algorithm', Sha256::class)
+
+        ->set('scheb_two_factor.trusted_jwt_encoder.configuration.key', InMemory::class)
+            ->factory([InMemory::class, 'plainText'])
             ->args(['%kernel.secret%'])
+
+        ->set('scheb_two_factor.trusted_jwt_encoder.configuration', Configuration::class)
+            ->factory([Configuration::class, 'forSymmetricSigner'])
+            ->args([
+                service('scheb_two_factor.trusted_jwt_encoder.configuration.algorithm'),
+                service('scheb_two_factor.trusted_jwt_encoder.configuration.key'),
+            ])
+
+        ->set('scheb_two_factor.trusted_jwt_encoder', JwtTokenEncoder::class)
+            ->args([service('scheb_two_factor.trusted_jwt_encoder.configuration')])
 
         ->set('scheb_two_factor.trusted_token_encoder', TrustedDeviceTokenEncoder::class)
             ->args([
--- a/src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
+++ b/src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
@@ -9,8 +9,6 @@ use Lcobucci\Clock\Clock;
 use Lcobucci\Clock\SystemClock;
 use Lcobucci\JWT\Configuration;
 use Lcobucci\JWT\Exception;
-use Lcobucci\JWT\Signer\Hmac\Sha256;
-use Lcobucci\JWT\Signer\Key\InMemory;
 use Lcobucci\JWT\Token\Plain;
 use Lcobucci\JWT\Validation\Constraint;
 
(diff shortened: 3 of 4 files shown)


src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
```diff
@@ -23,18 +23,17 @@ class JwtTokenEncoder
     public const CLAIM_FIREWALL = 'fwl';
     public const CLAIM_VERSION = 'vsn';
 
-    private Configuration $configuration;
+    private Clock $clock;
 
-    public function __construct(string $applicationSecret)
+    public function __construct(private Configuration $configuration, ?Clock $clock = null)
```
lcobucci, line 28: This is a BC-break... I allowed myself to do it because of the `@internal`

lcobucci:
btw if you decided to bump up the min dependency on `lcobucci/jwt`, you should also be able to use `UnencryptedToken` (interface) instead of `Plain`.

app/config/services.yaml
```diff
@@ -26,3 +26,9 @@ services:
 
     # add more service definitions when explicit configuration is needed
     # please note that last definitions always *replace* previous ones
+
+    scheb_two_factor.trusted_jwt_encoder.configuration.key:
```
lcobucci, line 30: I didn't want to change `scheb_two_factor.trusted_jwt_encoder.configuration.algorithm` instead to keep compatibility with older `lcobucci/jwt` versions :+1:

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
```diff
@@ -55,14 +54,18 @@ public function decodeToken(string $token): ?Plain
...
 
         return $token;
     }
+
+    /** @return iterable<int, Constraint> */
+    private function validationConstraints(): iterable
+    {
+        yield new Constraint\SignedWith($this->configuration->signer(), $this->configuration->signingKey());
+        yield new Constraint\ValidAt($this->clock); // replace with LooseValidAt once dependency on lcobucci/jwt is bumped up
+        yield from $this->configuration->validationConstraints();
```
lcobucci, line 69: I kept this here but it's very unlikely to have users adding validation constraints to the injected configuration (it's a possibility though, and custom claims are configurable by providing a different builder factory).

src/bundle/Resources/config/trusted_device.php
```diff
@@ -15,9 +18,22 @@
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
-        ->set('scheb_two_factor.trusted_jwt_encoder', JwtTokenEncoder::class)
+        ->set('scheb_two_factor.trusted_jwt_encoder.configuration.algorithm', Sha256::class)
+
+        ->set('scheb_two_factor.trusted_jwt_encoder.configuration.key', InMemory::class)
+            ->factory([InMemory::class, 'plainText'])
             ->args(['%kernel.secret%'])
```
scheb, line 25: I'd add an option to the bundle's configuration (defaulting to `%kernel.secret%`) to set the key for the trusted device token.

That would give people the ability to easily change the key for the trusted device token only, without changing the kernel.secret for the entire application and without overwrite the a service definition from the bundle (as you have demonstrated in [app/config/services.yaml](https://github.com/scheb/2fa/pull/158/files#diff-763476bab89e87b6546d922380ad2a190a43254a2ff04d178c55ab92b90ba813)). Overwriting services from 3rd party bundles just be the last resort.

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
```diff
@@ -23,18 +23,17 @@ class JwtTokenEncoder
...
-        $this->configuration = Configuration::forSymmetricSigner(new Sha256(), InMemory::plainText($applicationSecret));
-        $this->configuration->setValidationConstraints(new SignedWith($this->configuration->signer(), $this->configuration->signingKey()));
+        $this->clock = $clock ?? SystemClock::fromSystemTimezone();
     }
 
     public function generateToken(string $username, string $firewallName, int $version, DateTimeImmutable $validUntil): Plain
     {
         $builder = $this->configuration->builder()
-            ->issuedAt(new DateTimeImmutable())
+            ->issuedAt($this->clock->now())
```
scheb, line 36: What's the advantage of using `Clock` here, over using a DateTimeImmutable?

And wouldn't it be necessary to declare a dependency to `lcobucci/clock` in the composer.json, since we're directly referencing code from that component?

scheb (changes requested):
Thank you for the elaborate PR. I'm generally fine with the direction this is going. Just few additions/questions I have to clarify, before we can move forward.

src/bundle/Resources/config/trusted_device.php
lcobucci: Adding as configuration makes sense, the only thing is that we'd need to be able to distinguish b64 encoded chars and plain strings.

I'm saying that because `random_bytes(32)` is usually used to generate keys, then it needs to be b64 encoded to be used in config files.

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
lcobucci: https://blog.frankdejonge.nl/being-in-control-of-time-in-php/ has some info on the motivation for this. We're also working on the psr-20 to promote interoperability between different implementations.

I think that adding the dependency is up to your policy regarding direct/indirect dependencies. I'd probably add it if it was my library 👍

src/bundle/Resources/config/trusted_device.php
scheb: I wouldn't overcomplicate things here, I'd make it behave the same as `kernel.secret` => plain text value with minimum length of 32 characters (265 bit). People are obviously capable of configuring the `kernel.secret` value, so I'd expect them to be able doing the same for an (optional) JWT token value in the bundle configuration :D

lcobucci: I don't think it's overcomplicated... Based on [this](https://github.com/symfony/flex/blob/v2.2.3/src/Configurator/EnvConfigurator.php#L202) `kernel.secret` is a hexadecimal representation of random bytes...

Up to you, though 👍

lcobucci:
@scheb feel free to apply whichever changes you deem necessary here. My intention was to suggest **a way** and not **the way**.

Don't feel pushed into merging this PR 👍

src/trusted-device/Security/TwoFactor/Trusted/JwtTokenEncoder.php
scheb: Ok, I get the point of encapsulating the question "what time is now", which is really helpful for the matter of testing. ~Though in this specific context I don't really see the benefit of pulling in another component/dependency, especially since we're not using its capabilities in the tests. It could be a `DateTimeImmutable` and then I wouldn't need to deal with another dependency.~

~If a standardized clock interface is becoming a thing, I'd rather brush over the entire bundle once, to replace all time-dependencies with a clock implementation and have it aligned across the entire library.~

^ Keeping this to document my though process.

Just realized that `Constraint\ValidAt` is depending on `Lcobucci\Clock\Clock`, so my idea of "just use a `DateTimeImmutable`" is obviously not feasible. In that case: Since we're referencing components from `lcobucci/clock` in the 2fa-bundle code, imo it has to go into the `composer.json`, because I cannot necessarily rely on 3rd party packages to pull in the dependencies that I'm referring in my code.

src/bundle/Resources/config/trusted_device.php
scheb: I believe that's just their cheap way of making random bytes YAML-safe, when those config files are generated via Symfony Flex ;) I don't think that value is actually _required_ to be hexadecimal and decoded `hex2bin` in Symfony.

https://github.com/symfony/symfony/search?q=hex2bin Doesn't look like the secret being decoded, likely just used as raw bytes.

So anyways, I'd be totally fine with making that config option in the bundle also raw bytes.

scheb:
Alright, will take care of this, once I find the time

scheb:
[v6.3.0](https://github.com/scheb/2fa/releases/tag/v6.3.0) released including your commits. Thanks for your work!

PR #161, opened 2022-08-31
Fix return type in docs

benblub:
The TwoFactorInterface has a nullable return type.

```php
<?php

declare(strict_types=1);

namespace Scheb\TwoFactorBundle\Model\Totp;

interface TwoFactorInterface
{
    // ...

    /**
     * Return the configuration for TOTP authentication.
     */
    public function getTotpAuthenticationConfiguration(): ?TotpConfigurationInterface;
}

```

**Description**

### DIFF
--- a/doc/providers/totp.rst
+++ b/doc/providers/totp.rst
@@ -88,7 +88,7 @@ and the period of the temporary codes.
            return $this->username;
        }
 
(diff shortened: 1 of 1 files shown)


scheb:
Thanks for fixing this!

PR #165, opened 2022-11-04
Add optional ThrottlingListener to throttle two factor attempts

tostiheld:
**Description**
We recently had our platform pentested and one of the findings was that two factor codes were able to be brute forced by an attacker. In our case, quite easily in fact, because we are using TOTP with 6 digit codes. I added this listener to our core library, and now all our applications are protected from brute forcing the two factor codes. I thought it might be nice to (optionally) add this to this bundle. Let me know what you think. If you think it's a good idea, I will add a unit test.

### DIFF
--- a/src/bundle/DependencyInjection/Configuration.php
+++ b/src/bundle/DependencyInjection/Configuration.php
@@ -48,6 +48,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('ip_whitelist_provider')->defaultValue('scheb_two_factor.default_ip_whitelist_provider')->end()
                 ->scalarNode('two_factor_token_factory')->defaultValue('scheb_two_factor.default_token_factory')->end()
                 ->scalarNode('two_factor_condition')->defaultNull()->end()
+                ->scalarNode('rate_limiter')->defaultNull()->end()
             ->end();
 
         /** @psalm-suppress ArgumentTypeCoercion */
--- a/src/bundle/DependencyInjection/SchebTwoFactorExtension.php
+++ b/src/bundle/DependencyInjection/SchebTwoFactorExtension.php
@@ -4,9 +4,11 @@ declare(strict_types=1);
 
 namespace Scheb\TwoFactorBundle\DependencyInjection;
 
+use Scheb\TwoFactorBundle\Security\TwoFactor\Event\ThrottlingListener;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
@@ -67,6 +69,13 @@ class SchebTwoFactorExtension extends Extension
         }
 
         $this->configureBackupCodeManager($container, $config);
+
+        if (isset($config['rate_limiter'])) {
+            $container->register('scheb_two_factor.rate_limiter', ThrottlingListener::class)
+                ->setArguments([
+                    '$rateLimiterFactory' => new Reference($config['rate_limiter'])
+                ]);
+        }
     }
 
     /**
--- a/src/bundle/Security/TwoFactor/Event/ThrottlingListener.php
+++ b/src/bundle/Security/TwoFactor/Event/ThrottlingListener.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Event;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+use Symfony\Component\RateLimiter\RateLimiterFactory;
+
+final class ThrottlingListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private RateLimiterFactory $rateLimiterFactory
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            TwoFactorAuthenticationEvents::ATTEMPT => 'onTwoFactorAttempt',
+            TwoFactorAuthenticationEvents::SUCCESS => 'onTwoFactorSuccess',
+        ];
+    }
+
+    public function onTwoFactorAttempt(TwoFactorAuthenticationEvent $event): void
+    {
+        $rateLimiter = $this->rateLimiterFactory->create($event->getRequest()->getClientIp());
+
+        if (!$rateLimiter->consume()->isAccepted()) {
+            throw new TooManyRequestsHttpException();
+        }
+    }
+
+    public function onTwoFactorSuccess(TwoFactorAuthenticationEvent $event): void
+    {
+        $rateLimiter = $this->rateLimiterFactory->create($event->getRequest()->getClientIp());
+        $rateLimiter->reset();
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/mailer": "^5.4 || ^6.0",
         "symfony/yaml": "^5.4 || ^6.0",
-        "vimeo/psalm": "^4.0"
+        "vimeo/psalm": "^4.0",
+        "symfony/rate-limiter": "^5.4 || ^6.0"
     },
     "autoload": {
         "psr-4": {


wouterj:
Given this is a colleague of mine, I think this is a good idea :)

Just a little heads-up: Symfony 6.2 introduced peekable rate limiters in order to reduce load on the limiter/lock systems: https://github.com/symfony/symfony/pull/46110 It would be a good idea to use them in this listener for Symfony 6.2 applications (this can be checked using  `class_exists(\Symfony\Component\HttpFoundation\RateLimiter\PeekableRequestRateLimiterInterface::class)`).

scheb:
Could make sense to add this to the bundle. I have [this page](https://symfony.com/bundles/SchebTwoFactorBundle/current/brute_force_protection.html) in the documentation for ages, but I don't know how many people actually read and do it. I think it makes sense to encourage people to adopt good security practices by adding an out-of-the-box solution.

Two things that I'm currently concerned about:
- I'd make this optional, per default disabled. I don't want to introduce new hidden behavior, when people are upgrading to the new minor version. Could be default enabled with the next major release.
- I don't want to force another dependency onto people with `symfony/rate-limiter`, one that they might not even be using. So I'm wondering how to do that. Either make it another sub-package of the bundle to be installed, or _not_ making it a hard dependency and checking on container compilation - when the feature is enabled - if the component is present.

tostiheld:
~I see I made a mistake (I always register the listener, even though I intended for that to be optional).~ [Edit: upon further inspection I think I did it right the first time? Please let me know, I'm not super familiar with bundle configuration]

I agree with the 2 points you raised @scheb, and I tried to implement it like that. I added `symfony/rate-limiter` to dev-dependencies for that reason as well.

src/bundle/DependencyInjection/Configuration.php
```diff
@@ -48,6 +48,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->scalarNode('ip_whitelist_provider')->defaultValue('scheb_two_factor.default_ip_whitelist_provider')->end()
                 ->scalarNode('two_factor_token_factory')->defaultValue('scheb_two_factor.default_token_factory')->end()
                 ->scalarNode('two_factor_condition')->defaultNull()->end()
+                ->scalarNode('rate_limiter')->defaultNull()->end()
```
scheb, line 51: I was rather thinking about a true/false switch here. I wouldn't know what to put here ;)

src/bundle/DependencyInjection/SchebTwoFactorExtension.php
```diff
@@ -67,6 +69,13 @@ public function load(array $configs, ContainerBuilder $container): void
         }
 
         $this->configureBackupCodeManager($container, $config);
+
+        if (isset($config['rate_limiter'])) {
```
scheb, line 73: This should check if the rate-limiter component is present and if not throw a meaningful error message to let the user know, they have to install that components (to make it optional and only require people to install it, when it's actually needed).

src/bundle/Security/TwoFactor/Event/ThrottlingListener.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Event;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+use Symfony\Component\RateLimiter\RateLimiterFactory;
+
+final class ThrottlingListener implements EventSubscriberInterface
```
scheb, line 11: Please remove the `final` keyword and add PHPDoc `@final` (check other classes for reference). That's how I do it in the library to keep myself the ability to mock things if necessary.

src/bundle/Security/TwoFactor/Event/ThrottlingListener.php
```diff
@@ -0,0 +1,40 @@
...
+        $rateLimiter = $this->rateLimiterFactory->create($event->getRequest()->getClientIp());
+
+        if (!$rateLimiter->consume()->isAccepted()) {
+            throw new TooManyRequestsHttpException();
+        }
+    }
+
+    public function onTwoFactorSuccess(TwoFactorAuthenticationEvent $event): void
+    {
+        $rateLimiter = $this->rateLimiterFactory->create($event->getRequest()->getClientIp());
```
scheb, line 37: This should have some kind of unique keyword in addition to the IP, to avoid conflicts with other pieces of code, e.g.:

```php
private const RATE_LIMITER_KEY_PREFIX = "scheb_two_factor.brute_force_protection.";

$this->rateLimiterFactory->create(self::RATE_LIMITER_KEY_PREFIX  . $event->getRequest()->getClientIp())
```

scheb (changes requested):
Necessary, so that I can consider this for merge:

- Fix code style issues
- Documentation for the new configuration setting
- Test coverage

src/bundle/DependencyInjection/Configuration.php
tostiheld: I intended for a rate limiter service to be put here, since those are highly configurable and I don't want to decide for the user what kind of rate limiter to use.

src/bundle/DependencyInjection/SchebTwoFactorExtension.php
tostiheld: I can add a check for the component too. Does that mean you want to remove `symfony/rate-limiter` from the dev dependencies? We need to check if the key is set as well, because we need it to inject the right rate limiter service.

src/bundle/Security/TwoFactor/Event/ThrottlingListener.php
tostiheld: I can do that of course, but I'm not sure if it's necessary, because the rate limiter component will automatically prefix the key with its ID (see https://github.com/symfony/rate-limiter/blob/6.2/RateLimiterFactory.php#L46)

src/bundle/DependencyInjection/Configuration.php
scheb: Could you give me an example how to use this? How would such a RateLimiterFactory service be configured?

Could be that I didn't fully get it and my comments don't make much sense 🙈

tostiheld: Say for example you have this in your `rate_limiter.yaml`:

```
framework:
    rate_limiter:
        mfa:
            policy: 'token_bucket'
            limit: 4
            rate:
                interval: '5 minutes'
                amount: 1
            cache_pool: 'some.cache.pool'
            lock_factory: 'some.lock.factory'
```

Then in `scheb_2fa.yaml` you can have something like:
```
scheb_two_factor:
    rate_limiter: 'limiter.mfa'
```

Now you will have your MFA attempts rate limited via the 'token bucket' strategy, with max 4 attempts and 1 new attempt every 5 minutes. The component documentation does a much better job at explaining how it works and how the different limiting strategies work: https://symfony.com/doc/current/rate_limiter.html.

scheb: Ahh, I see. I wasn't ware you have all these configurations options on the rate limiter. Makes much more sense now!

Under these conditions we can keep everything as-is. _Though_, we need to have good defaults for the rate limiter in the documentation, as most people would just go with it.

src/bundle/DependencyInjection/SchebTwoFactorExtension.php
scheb: Resolved via https://github.com/scheb/2fa/pull/165#discussion_r1038863299, no need to change.

src/bundle/Security/TwoFactor/Event/ThrottlingListener.php
scheb: Resolved via https://github.com/scheb/2fa/pull/165#discussion_r1038863299, no need to change.

hdk-pd:
I'm reading up on Symfony a bit and looking at some code to dig into the concepts and ecosystem and came across this, so I'm sorry if I am a bit off in some technical details.

The throttle addresses `TwoFactorAuthenticationEvents::ATTEMPT` events, in plain words: a successful login attempt with correct account credentials and the rate limit key is the request IP, right? So I think the threat model here is someone trying to gain access to a specific account or a specific set of accounts. I don't think the usual automated credential sprayers that are doing this at scale will continue when faced with a 2FA authentication - generally speaking.
My suggestion would be to use the users identifier instead of the request IP as rate limit key. Checking the event and the TokenInterface, I think this should be what I'm suggesting:

```php
$rateLimiter = $this->rateLimiterFactory->create($event->getToken()->getUserIdentifier());
```

This highly depends on the rate limiter and code generator configuration and implementation. Given a relatively short 2FA code and simple code (i.e. 6 digit) and some attempts available, access to accounts could possibly be gained with several cloud providers, proxy services etc. ([Example bug bounty writeup for an Instagram 2FA brute force rate limit issue](https://thezerohack.com/hack-any-instagram))
With my suggestion, an attacker with a single IP might have more attempts in total, but all on different accounts with different 2FA codes to brute force. However, an attacker looking for a specific account can not get around the rate limit key with different IPs. Every user will have a specific set of attempts. The worst scenario I could imagine is an attacker creating a DoS scenario for a victims account by blocking the 2FA functionality with ongoing attempts. Given this requiring the attacker to know the users credentials, this might be an acceptable risk and can be fixed with a password reset that is overdue anyways. Privacy enthusiasts behind a VPN or Tor might be blocked by the IP key, too.
Please let me know what you think.

scheb:
Closing due to inactivity.

PR #168, opened 2022-11-28
Check if the session is started

trsteel88:
For bug fixes:
 - Avoid starting session usage

The Symfony Framework currently detects usage on the session. If it is used, Symfony will output private headers. See https://github.com/symfony/symfony/blob/c79d4ab7e787eb27d83581ccb8e15a4e11e88df3/src/Symfony/Component/HttpKernel/EventListener/AbstractSessionListener.php#L197

This PR updates the Form listener to check if the session isStarted() before calling getToken on the token storage. Doing so, means the session usage counter isn't incremented.

Alternatively, the check for getToken() should be fired after isAuthFormRequest().

We currently have a site where we're trying to cache the front-end. However, this listener is firing for the front-end firewall even though it isn't relevant.

### DIFF
--- a/src/bundle/Security/TwoFactor/Event/TwoFactorFormListener.php
+++ b/src/bundle/Security/TwoFactor/Event/TwoFactorFormListener.php
@@ -30,10 +30,8 @@ class TwoFactorFormListener implements EventSubscriberInterface
         if (!$request->hasSession()) {
             return;
         }
-        
-        $session = $request->getSession();
 
-        if (!$session->isStarted()) {
+        if (!$this->twoFactorFirewallConfig->isAuthFormRequest($request)) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


trsteel88:
@scheb any change of this being merged?

scheb:
I can see a weird edge-case:

- When there is a session
- the session is not "started"
- the page shown is the 2fa auth form page
- the event wouldn't fire, but it should.

So, the correct approach would be to reorder to conditions. Check `isAuthForm` first, then - and only then - check the token. So that on all pages, except the 2fa auth form page, the listener would not touch the session.

scheb:
This is what I mean:

```php
    public function onKernelRequest(RequestEvent $requestEvent): void
    {
        $request = $requestEvent->getRequest();
        if (!$request->hasSession()) {
            return;
        }

        if (!$this->twoFactorFirewallConfig->isAuthFormRequest($request)) {
            return;
        }

        $token = $this->tokenStorage->getToken();
        if (!($token instanceof TwoFactorTokenInterface)) {
            return;
        }

        $event = new TwoFactorAuthenticationEvent($request, $token);
        $this->eventDispatcher->dispatch($event, TwoFactorAuthenticationEvents::FORM);
    }

```

trsteel88:
Thanks @scheb. I have updated that.

PR #171, opened 2023-01-19
Allow lcobucci/clock v3

IonBazan:
**Description**

Simply allows https://github.com/lcobucci/clock/releases/tag/3.0.0 - which switches from `stella-maris/clock` to `psr/clock`.

### DIFF
--- a/app/composer.json
+++ b/app/composer.json
@@ -4,11 +4,12 @@
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "ext-json": "*",
+        "doctrine/annotations": "^1.13",
         "doctrine/doctrine-bundle": "^2.0.3",
         "doctrine/orm": "^2.0",
         "egulias/email-validator": "^3.0",
         "endroid/qr-code": "^4.0",
-        "lcobucci/clock": "^2.0",
+        "lcobucci/clock": "^2.0 || ^3.0",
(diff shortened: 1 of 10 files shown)


IonBazan:
Test failures seem unrelated 🤷🏻‍♂️

maxhelias:
I think just add `doctrine/annotations` in the `require-dev` section should fix the test

IonBazan:
Thanks @maxhelias - it did the trick!

IonBazan:
OK I went a bit further and switched to use `attribute` driver for both routes and entity mapping. This way we can get rid of `doctrine/annotations` completely as we already require PHP 8+ and Symfony 5.4+ 🚀

maxhelias:
In this case you can also delete the `app/config/routes/annotations.yaml` file or alternatively a `composer recipes:update` should tell you what to do.

IonBazan:
Removing `app/config/routes/annotations.yaml` is only possible on Symfony 6. Since the app allows 5.4, we need to keep it there for BC.

scheb:
Wow, thanks for going through the effort of updating the thing with the annotations as well! 🤩

Discovered that issue with the doctrine bundle 2.8 just a few days earlier in another project, but it took me way longer to figure out what's going on 😆

Will have a look at the PR on the weekend!

scheb:
Released as v6.5.0

Thanks for the clean commit history! 🙇

PR #172, opened 2023-01-26
fix getContentType deprecation since symfony 6.2

jeremywaguet:
**Description**
Handle both legacy Request::getContentType() and Request::getContentTypeFormat() introduced with Symfony 6.2


scheb:
Thanks. Released as v6.6.0

PR #176, opened 2023-02-17
Trivial corrections to sample JS in API integration docs

danielrhodeswarp:
**Description**
Trivial corrections to sample JS in API integration docs. Makes it easier to copy-and-paste :-)

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -82,7 +82,7 @@ To implement such a response you need to a custom success handler:
        {
            if ($token instanceof TwoFactorTokenInterface) {
                // Return the response to tell the client two-factor authentication is required.
-               return new Response('{"login": "success": "two_factor_complete": false}');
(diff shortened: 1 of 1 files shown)


scheb:
I can see that the colon is obviously a syntax error. Is there any reason why you changed the name of the field?

danielrhodeswarp:
Hello Christian,
Thanks for getting back - it's a very interesting package.
Yes, so JavaScript - like a lot of languages - doesn't allow a variable name or identifier to start with a digit.
I was getting errors with my front-end when using the sample code as-is.
This change is a small suggestion to make the sample code easier to copy and paste :-)
Kind regards,Daniel

Sent from Yahoo Mail on Android 
 
  On Thu, 23 Feb 2023 at 17:44, Christian ***@***.***> wrote:   


I can see that the colon is obviously a syntax error. Is there any reason why you changed the name of the field?

PR #179, opened 2023-02-26
Add example attributes on documentation

maxhelias:
I'm not sure if i use the right `configuration-bloc`, maybe @javiereguiluz can review it, please ?

### DIFF
--- a/doc/providers/totp.rst
+++ b/doc/providers/totp.rst
@@ -57,43 +57,81 @@ and the period of the temporary codes.
     Be warned, custom configurations will not be compatible with the defaults of Google Authenticator app any more. You
     will have to use another application (e.g. FreeOTP on Android).
 
-.. code-block:: php
+.. configuration-block::
 
-   <?php
+    .. code-block:: php
 
-   namespace Acme\Demo\Entity;
+       <?php
 
-   use Doctrine\ORM\Mapping as ORM;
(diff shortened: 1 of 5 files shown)


doc/providers/totp.rst
```diff
@@ -57,43 +57,80 @@ and the period of the temporary codes.
     Be warned, custom configurations will not be compatible with the defaults of Google Authenticator app any more. You
     will have to use another application (e.g. FreeOTP on Android).
 
-.. code-block:: php
-
-   <?php
+.. configuration-block::
+    .. code-block:: php
```
javiereguiluz, line 62: Here, let's add a blank line between `configuration-block` and `clode-block` directives.

javiereguiluz (commented):
Yes, I think this is correct. The required RST directive structure is:

```rst
.. configuration-block::

    .. code-block:: php-annotations

        // code here...

    .. code-block:: php-attributes

        // code here...
```

doc/providers/totp.rst
maxhelias: Done! Thanks for the feedback

scheb:
Stupid question: What is this doing compared to the previous implementation?

javiereguiluz:
@scheb it creates a clickable set of tabs where readers can select their preferred format (and that format is persisted on the site):

scheb:
Ok, got it 👍 Sweet!

PR #180, opened 2023-03-03
Allow lcobucci/jwt v5

maxhelias:


scheb:
Why did you lower the PHP version on the CI job?

maxhelias:
This is to fix psalm which automatically target to a php 8.0 scan but with php 8.1 installed packages. And some libs implement these new features that are not well understood by psalm

scheb:
Sorry, I totally forgot about this PR. Merged and released as v6.8.0.

PR #181, opened 2023-03-03
Remove incorrect @ for PHP 8 ORM attribute

rskuipers:


scheb:
Thanks for pointing that out. Figured there are some more places in the documentation where this is wrong and fixed that as well.

rskuipers:
@scheb Ah, I probably could've checked that for you as well. Thanks for merging and your OSS efforts 👍

PR #187, opened 2023-05-17
Add subject_email  parameter

sdespont:
This PR brings the possibility to define the subject of the mail sent with the authentication code.
The parameter `subject_email` have been added, with a default value equals to the actual text "Authentication Code"

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ jobs:
             fail-fast: false
             matrix:
                 php-version:
-                    - '8.1'
+                    - '8.0'
 
         steps:
             -   name: 'Checkout code'
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ composer.lock
 /app/var/
 /app/vendor/
 /app/symfony.lock
+# PHPStorm
+.idea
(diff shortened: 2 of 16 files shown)


sdespont:
@scheb what do you think about having a configurable prefix text before the code in the email? I don't want to spend time to push a PR if you don't agree with this approach

scheb:
Hmm, wouldn't want to add more configuration options. I'm not having the intention to provide the ultimate configurable mailer with the bundle.

Way to go would rather be to implement a custom mailer, which generates the email exactly as you want it. The on-board mailer is just doing the bare minimum to have something working.

sdespont:
@scheb you are right, implement a custom mailer is easy https://symfony.com/bundles/SchebTwoFactorBundle/current/providers/email.html#custom-mailer

After reflection, suffix or prefix is not a good idea. But permitting to set the email subject of the default email still makes sense for me.

scheb:
There seem a whole lot of changes in the diff, that aren't related to this merge request. In its current form I cannot merge it, as it will mess things up.

 If you want to proceed, please review the diff and provide a cleaned-up changeset that just contains your changes.

scheb:
Closing due to inactivity. If you feel like follow-up on this, please create another PR. Thank you!

PR #192, opened 2023-06-20
Improve static analysis

norkunas:
**Description**
`OTPInterface::verify` expects `$otp` parameter as non-empty-string and also it's `createFromSecret`, `getSecret` and `setSecret` methods contracts are non-empty-string


scheb:
Fair enough, though now the Psalm check is failing with those changes. As long as these aren't handled, I can't merge this.

If we're annotating `TotpAuthenticatorInterface` that way because of OTPInterface, then the same thing should be done for  `GoogleAuthenticatorInterface`, since it's having the same dependency.

norkunas:
Yeah, when I saw the CI, then I started to add those in other places, but there are so many, just stopped a little bit

scheb:
If you consider follow-up on this, please create a new PR. Until then, I'm closing this one.

PR #197, opened 2023-08-03
chore: add ci test-runs for php 8.3

Chris53897:
I know it is a bit early. But tests look good a the moment.


scheb:
Thanks for the effort. Though, the way I handle it in this library is, that future releases like PHP 8.3 will only be officially supported, once there is a stable release and I can be sure it's working.

For the time being, I've taken over your changes to a development branch [6.x-php8.3](https://github.com/scheb/2fa/tree/6.x-php8.3). Thanks for contributing!

PR #198, opened 2023-08-03
docs: fix non-breaking space

Chris53897:
change the non-breaking space (NBSP) to a normal space


scheb:
Why would you want to change that? The non-breaking space is intentional, because I don't want the term "Symfony 6" to be torn apart by a line break.

Chris53897:
Sorry, i thought this was unintentional

Issue #208, opened 2023-11-03
Validate code before actual Google Authentication / Different BruteForce protection compared to documentation


hracik:
**Description**

At the moment when user submits empty field or the length isn't the same as expected - default is 6, or typed letters it is validated by against the actual code which seems slower than validating with NotBlank or Lenght and only after that validate the actual code.

Bundle says in [documentation ](https://symfony.com/bundles/SchebTwoFactorBundle/6.x/brute_force_protection.html) that brute force is not enabled by default and some listeners are needed, but I did not write any and after I submitted a couple of empty code I got error Too many failed login attempts, please try again in 12 minutes and a bit later Too many failed login attempts, please try again in 50 minutes from symfony/security.

Seem like login_throttling in the firewall is doing the job.
```
login_throttling:
max_attempts: 3
interval: '60 minute'
```

My firewall 
```
firewalls:
        main:
            login_throttling:
                max_attempts: 3
                interval: '60 minute'
            lazy: true
            provider: app_user_provider
            logout:
                path: app_logout
            switch_user: true
            remember_me:
                secret: '%kernel.secret%'
                lifetime: 604800 # 1 week in seconds
                path: /
            user_checker: App\Security\UserChecker
            custom_authenticators:
                - App\Security\LoginFormAuthenticator
            two_factor:
                auth_form_path: 2fa_login    # The route name you have used in the routes.yaml
                check_path: 2fa_login_check  # The route name you have used in the routes.yaml
                enable_csrf: true
```

scheb:
That's a great finding! Didn't know `login_throttling` could do that. Added it to the documentation as the recommended method to add brute-force protection.

PR #211, opened 2023-12-09
Add ThrottlingListener to allow MFA attempt throttling

tostiheld:
Trying again after I forgot about #165 :smile: 

Still in draft, because I'm still working on tests, but I thought I would already put something up because of #SymfonyHackday

This basically does the same thing as the previous PR, with one notable improvement. I agree with the point that @hdk-pd made previously (https://github.com/scheb/2fa/pull/165#issuecomment-1376546058). Therefore I took some inspiration from https://github.com/symfony/symfony/pull/38308 and implemented a combined ratelimiter instead of a single one. It will now ratelimit on both IP and user identifier.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,8 @@
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/mailer": "^6.4 || ^7.0",
         "symfony/yaml": "^6.4 || ^7.0",
-        "vimeo/psalm": "^5.0"
+        "vimeo/psalm": "^5.0",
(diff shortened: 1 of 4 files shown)


scheb:
Thanks for the effort. We recently figured in https://github.com/scheb/2fa/issues/208 that the `login_throttling` feature on the firewall-level adds this behavior out-of-the-box, not only for the initial login, but also for the 2fa.

tostiheld:
Okay, I will close this.

Issue #212, opened 2023-12-31
Pass TwoFactorTokenInterface to PreferredProviderInterface


danielburger1337:
It would be helpful to pass the current token context to the `PreferredProviderInterface::getPreferredTwoFactorProvider()` method so that we can make decisions on what provider should be preferred when the preferred one is unavailable.

My use case is that I have a custom passkey provider and I also allow users to sign in with a passkey. Through a token attribute which I check in my TwoFactorProvider, I disable passkey support for users that originally authenticated with the passkey.

When `passkey` is the users preferred two factor provider and now the provider is blocked, the default provider order is used. 

This is kinda suboptimal because "email" is before "totp" and will send an email on request. It would be better to then fallback to the TOTP provider so that a potentially unnecessary email is not sent. 

An alternative would be to somehow allow to configure the order of providers, but that would require an extensive change to the codebase.

scheb:
Can't just change this, because it's a breaking change.

In my opinion, the interface should stay as simple as possible, since it's meant to be implemented in the user entity and return the user's preference. It's not really meant to contain a lot of complex business code.

The better solution would probably be to extract this piece of code https://github.com/scheb/2fa/blob/7990826108680acb90ea308d8115c62fe2a0cbc6/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php#L58-L75 into a service, which can be overwritten via config for the case when you'd need to inject your own custom logic.

PR #215, opened 2024-01-07
Add TwoFactorProviderDecider

danielburger1337:
See #212

Every scenario **should** have a dedicated test case.
I would appreciate it if we could merge and release this feature ASAP.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -100,7 +100,7 @@ Bundle Configuration
        # Must implement Scheb\TwoFactorBundle\Security\TwoFactor\Condition\TwoFactorConditionInterface
        two_factor_condition: acme.custom_two_factor_condition
 
-       # If you need custom conditions to decide what two factor provider to preferr.
+       # If you need custom conditions to decide what two factor provider to prefer.
        # Must implement  Scheb\TwoFactorBundle\Security\TwoFactor\Provider\TwoFactorProviderDeciderInterface
        two_factor_provider_decider: acme.custom_two_factor_provider_decider
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -100,6 +100,10 @@ Bundle Configuration
        # Must implement Scheb\TwoFactorBundle\Security\TwoFactor\Condition\TwoFactorConditionInterface
        two_factor_condition: acme.custom_two_factor_condition
 
+       # If you need custom conditions to decide what two factor provider to preferr.
+       # Must implement  Scheb\TwoFactorBundle\Security\TwoFactor\Provider\TwoFactorProviderDeciderInterface
+       two_factor_provider_decider: acme.custom_two_factor_provider_decider
+
 Firewall Configuration
 ----------------------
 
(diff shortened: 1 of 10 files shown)


scheb:
Thanks for handing this in. On the first view, this is looking quite good. I'll look into this more in detail, once I find time to.

doc/configuration.rst
```diff
@@ -100,6 +100,10 @@ Bundle Configuration
        # Must implement Scheb\TwoFactorBundle\Security\TwoFactor\Condition\TwoFactorConditionInterface
        two_factor_condition: acme.custom_two_factor_condition
 
+       # If you need custom conditions to decide what two factor provider to preferr.
```
scheb: Typo ;)
```suggestion
       # If you need custom conditions to decide what two factor provider to prefer.
```

src/bundle/Security/TwoFactor/Provider/TwoFactorProviderDeciderInterface.php
```diff
@@ -0,0 +1,17 @@
...
+use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenInterface;
+
+interface TwoFactorProviderDeciderInterface
+{
+    /**
+     * Return the alias of the preferred two-factor provider.
+     *
+     * @param string[] $activeProviders
+     */
+    public function getPreferredTwoFactorProvider(array $activeProviders, TwoFactorTokenInterface $token, object $user): string|null;
```
scheb: I'd suggest to adjust the interface slightly. Instead of passing the $user, better pass a `AuthenticationContextInterface $context` here, which provides pretty much all relevant information to make a decision about the preferred provider.

tests/Security/TwoFactor/Provider/TwoFactorProviderInitiatorTest.php
```diff
@@ -36,7 +37,7 @@ protected function setUp(): void
 
         $this->twoFactorTokenFactory = $this->createMock(TwoFactorTokenFactory::class);
 
-        $this->initiator = new TwoFactorProviderInitiator($providerRegistry, $this->twoFactorTokenFactory);
+        $this->initiator = new TwoFactorProviderInitiator($providerRegistry, $this->twoFactorTokenFactory, new TwoFactorProviderDecider());
```
scheb: This should be mocked to not create a dependency of the test towards the specific implementation of `TwoFactorProviderDecider`.

scheb (changes requested):
Excellent PR! 👍 

I have not much to change on that. I'd change the interface of the decider slightly, other than that, the approach looks really good.

tests/Security/TwoFactor/Provider/TwoFactorProviderInitiatorTest.php
danielburger1337: I hope the new test is the way you like it.

src/bundle/Security/TwoFactor/Provider/TwoFactorProviderDeciderInterface.php
danielburger1337: Thats a fair point. There really is no reason to not pass it along.

scheb:
Excellent PR, thank you! Merged and released as v7.1.0

PR #217, opened 2024-01-19
Nested IP whitelist

danielburger1337:
**Description**

This PR introduces support for array values in the `scheb_two_factor.ip_whitelist` configuration option. This allows the user to specify the new-ish [IpUtils::PRIVATE_SUBNETS](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpFoundation/IpUtils.php) constant together with some public IPs.

```yaml
scheb_two_factor:
    # This already worked
    ip_whitelist: !php/const Symfony\Component\HttpFoundation\IpUtils::PRIVATE_SUBNETS
```

Now this works as well:

```yaml
scheb_two_factor:
    ip_whitelist:
        - 9.9.9.9
        - [1.1.1.1, 1.0.0.1] 
        - !php/const Symfony\Component\HttpFoundation\IpUtils::PRIVATE_SUBNETS

### DIFF
--- a/src/bundle/DependencyInjection/Configuration.php
+++ b/src/bundle/DependencyInjection/Configuration.php
@@ -15,6 +15,7 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
 use Symfony\Component\Security\Http\Authenticator\Token\PostAuthenticationToken;
 use function interface_exists;
+use function is_iterable;
 
 /**
  * @final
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/bundle/DependencyInjection/Configuration.php [hunk 1]
```diff
@@ -42,6 +43,17 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                    ->beforeNormalization()
+                        ->ifArray()
+                        ->then(static function (mixed $value): array {
+                            $values = [];
+                            foreach (self::flatten($value) as $v) {
+                                $values[] = $v;
+                            }
+
+                            return $values;
+                        })
```
scheb: I believe this could be simplified with PHP on-board functions:

```
return iterator_to_array(new RecursiveIteratorIterator(new RecursiveArrayIterator($value)), false);
```

danielburger1337: Funny. I have actually used this in the past and completely forgot that it existed.

scheb:
Thanks for contributing once again! 🙇

Released as v7.2.0

PR #219, opened 2024-01-25
Update documentation - Response is required

RichardJong:
Match AuthenticationFailureHandlerInterface - Response is required


scheb:
Thanks for fixing it!

PR #228, opened 2024-04-19
introduce CacheableVoterInterface to reduce amounts of calls

### LINKED ISSUE #227: Increase performance by using CacheableVoterInterface
**Description**

Symfony 5.4 introduced cacheable Voters, which speed up apps, which use a lot of `is_granted` calls.
You can define which attributes the Voter supports, and thus it will not be called for other attributes.

- https://symfony.com/blog/new-in-symfony-5-4-faster-security-voters
- https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Security/Core/Authorization/Voter/CacheableVoterInterface.php

In one of my apps, I opened the profiler for a listing page, which triggers around 1000 `is_granted` calls.
Each of them calls `Scheb\TwoFactorBundle\Security\Authorization\Voter\TwoFactorInProgressVoter` in addition to the specific app Voter, see this example screenshot:
Bildschirmfoto 2024-04-17 um 21 35 00

By changing the voter to this 
```php

class TwoFactorInProgressVoter implements CacheableVoterInterface
{

    public function supportsAttribute(string $attribute): bool
    {
        return $attribute === self::IS_AUTHENTICATED_2FA_IN_PROGRESS || $attribute === AuthenticatedVoter::PUBLIC_ACCESS;
    }

    public function supportsType(string $subjectType): bool
    {
        return $subjectType === 'null';
    }

    ... 
}
````

I was able to remove the Voter from almost all of the 1k `is_granted` decisions:

Bildschirmfoto 2024-04-17 um 21 45 31

Which reduced the time from this:

Bildschirmfoto 2024-04-17 um 21 47 11

to this:

Bildschirmfoto 2024-04-17 um 21 47 35

A pretty low hanging fruit, which I'd like to address in a PR if that's okay!? 
Or is there anything I am missing?

kevinpapst:
Closes #227

**Description**

See #227 as well.

Adds the `Symfony\Component\Security\Core\Authorization\Voter\CacheableVoterInterface` which dramatically reduces the amount of calls to the 2FA specific Voter.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -11,6 +11,7 @@ $rules = [
     'phpdoc_separation' => ['groups' => [['test', 'dataProvider']]],
     'nullable_type_declaration_for_default_null_value' => true,
     'no_superfluous_phpdoc_tags' => ['remove_inheritdoc' => false],
+    'nullable_type_declaration' => false,
 ];
 
 $finder = PhpCsFixer\Finder::create()
--- a/src/bundle/Security/Authorization/Voter/TwoFactorInProgressVoter.php
+++ b/src/bundle/Security/Authorization/Voter/TwoFactorInProgressVoter.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Scheb\TwoFactorBundle\Security\Authorization\Voter;
 
 use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenInterface;
-use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Authorization\Voter\AuthenticatedVoter;
 use Symfony\Component\Security\Core\Authorization\Voter\CacheableVoterInterface;
@@ -47,6 +46,6 @@ class TwoFactorInProgressVoter implements CacheableVoterInterface
 
     public function supportsType(string $subjectType): bool
     {
-        return $subjectType === 'null' || $subjectType === Request::class;
+        return true;
     }
 }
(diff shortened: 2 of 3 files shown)


scheb:
Looks like the integration tests really don't like it and seem to be stuck in an endless loop.

kevinpapst:
I added code like that to more than a dozen Voters across multiple projects 🤷 
I don't think that it has to do with my change, but who knows .... 

Can you restart them and if that doesn't help: how can I run them locally?

scheb:
Restarted, but it looks stuck again.

You can run the integration tests locally from the `app` folder. So basically `cd` into `app`, run `composer install` and then `vendor/bin/phpunit`.

See: https://github.com/scheb/2fa/tree/7.x/app

kevinpapst:
Ups, missed the app directory, thanks. Will test and report back!

kevinpapst:
Okay, the test showed thar the Voter receives a Request object during login as type and therefor it was stuck in a loop.

I am not familiar enough with the internals of the security system, so I don't know if any other type might be valid as well, such as the User or Token. Is it safe to only accept `null` and `Request`? We could also return true for every type and let the cache only work on the two attribute names.

Should make no big difference performance wise, as attributes are checked first anyway:
https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Security/Core/Authorization/AccessDecisionManager.php#L102

What do you say?

scheb:
The `$subject` is completely irrelevant to the voter, so I believe it should just return `true`, because the subject can be anything. That's what Symfony does for its own cached voters, such as the https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Security/Core/Authorization/Voter/AuthenticatedVoter.php#L88-L103, which is also not doing anything with the subject.

kevinpapst:
Agree 👍 changed. Thanks for you input.

scheb:
Nice! Would you please rebase onto to target branch. I just solved these coding standards errors. So we should get a proper result for your PR then.

kevinpapst:
done

scheb:
Thank you for contributing! 🙇 

Released as v7.3.0.

PR #235, opened 2024-06-18
Wire clock for totp factories

norkunas:
**Description**
Not passing the clock is deprecated


scheb:
Thanks, released as v7.4.0

PR #236, opened 2024-07-03
Indonesian translation

priyadi:
Adds Indonesian translation.


scheb:
Awesome, thanks! Released as v7.5.0

PR #245, opened 2024-09-05
Adding an event informing system that Email AuthCode has been validated

mdzwigaladp:
**New Feature:** Dispatching an Event when Email AuthCode is successfully validated.

**Motivation:** In the use case we are working on, it would be useful to subscribe to an event that would inform our component that the code that has been sent via an email has been successfully validated.

**Why it's a good idea to solve it this way?**: Using events is a common practice to solving this kind of problems while maintaining proper separation of concerns and obeying open-closed principle. This event doesn't contain any excess data: Just an email of the validated user and a descriptive name, that's it.

### DIFF
--- a/src/bundle/Resources/config/two_factor_provider_email.php
+++ b/src/bundle/Resources/config/two_factor_provider_email.php
@@ -39,6 +39,7 @@ return static function (ContainerConfigurator $container): void {
             ->args([
                 service('scheb_two_factor.security.email.code_generator'),
                 service('scheb_two_factor.security.email.form_renderer'),
+                service('event_dispatcher'),
             ])
 
         ->alias(CodeGeneratorInterface::class, 'scheb_two_factor.security.email.code_generator')
--- a/src/email/Event/EmailCodeValidated.php
+++ b/src/email/Event/EmailCodeValidated.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Event;
+
+readonly class EmailCodeValidated
+{
+    public function __construct(
+        public string $email,
+    ) {
+    }
+}
--- a/src/email/Event/EmailTwoFactorEvents.php
+++ b/src/email/Event/EmailTwoFactorEvents.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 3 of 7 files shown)


chaours:
I would like make the same when it's a Backup Code that has been used. Thanks for the example :) I will do a PR like yours.

scheb:
If we add this for the email provider, it would make sense to add it for the Google Authenticator and TOTP provider as well.

But instead of adding an event for each provider, I'd rather add a generic event for code validation, which is dispatched in https://github.com/scheb/2fa/blob/7.x/src/bundle/Security/Http/EventListener/CheckTwoFactorCodeListener.php

For a backup code event, it should be its own event and dispatched here:
https://github.com/scheb/2fa/blob/7.x/src/backup-code/Security/Http/EventListener/CheckBackupCodeListener.php

chaours:
Thanks for your tips. I have to fork the repo before send you a PR ?

scheb:
Yes, for making a PR you have to fork first.

chaours:
Ok. I will do it in hours. Thanks !

mdzwigaladp:
Hey @scheb thank you for your input!
I like the idea to adding this kind of event into the other providers.

About the generic events - I think that it would be better to keep them as a separate events. Various type of providers may contain different metadata (let's say - sms can have information about phone number that can be used) and with explicit events we can have explicit methods providing this information and making use of strong typing - instead of relying for example on some kind of `metadata` non-typed array. These would be provider-specific events, giving more intel about the process, not like a 'core' event - which are already there.

WDYT?

scheb:
Okay, fair enough, I see the point with different metadata. So from my end, we could go with different event classes. Under these conditions it also makes sense to put the code into the `validateAuthenticationCode` method, since this is the method all providers have to check their codes.

My main request would be, that we do it for all 2fa providers and not just for a single one :)

In general, I'd add `$user` and `$code` as event properties, since these are coming as parameters of the `validateAuthenticationCode` method, and adding the code would make it is possible to check which code specifically was consumed.

I'd lean towards having an abstract event class on the bundle level, that comes with the default properties for user and code. Then providers can extend from that it to add their own additional properties.

scheb:
Closing this unfinished PR. Such events have been added in version 7.8.0.

PR #252, opened 2024-10-06
Feat / Creation of BackupCodeUsedEvent

chaours:
**Description**
This PR let us have a BackupCodeUsed event dispatching.
We wanted to send a addFlash message to inform that the backup code is now obsolete and highly recommand to generate a new one right now.
I hope this PR will interest you.
Thanks

### DIFF
--- a/doc/events.rst
+++ b/doc/events.rst
@@ -28,6 +28,13 @@ Constant: ``Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorAuthenticati
 
 Is dispatched when two-factor authentication is attempted, right before checking the code.
 
+``scheb_two_factor.authentication.backup_code_used``
+-------------------------------------------
+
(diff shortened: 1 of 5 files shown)


chaours:
Hum... Many errors... I can try fix them.

scheb:
There seems to be an issue with the unit test.

I also think, the implementation is a bit convoluted, as it introduces two extra dependencies into the `CheckBackupCodeListener` class, just to dispatch that event (the token at least could be taken from the passport object).

Though, `TwoFactorAuthenticationEvents` might not be the right event class for this. `TwoFactorAuthenticationEvents` was originally meant to represent the different phases of the 2fa process, so I think it's fair, if we'd introduce a dedicate event class for "checking codes". Could take the token (which contains information about the user) and the code as arguments.

scheb:
Closing this unfinished PR. Such events have been added in version 7.8.0.

PR #253, opened 2024-10-18
Allow tests to run on PHP8.4

heiglandreas:
This then also prepares the package to be released for PHP8.4 runtimes

### DIFF
--- a/.github/actions/setup-build/action.yaml
+++ b/.github/actions/setup-build/action.yaml
@@ -66,6 +66,10 @@ runs:
             shell: bash
             run: composer require ${{ inputs.composer-require }} --dev --no-update --working-dir=${{ inputs.composer-working-dir }}
 
+        -   name: "Remove Psalm from dependencies"
+            shell: bash
+            run: composer remove --dev vimeo/psalm
+
         -   name: "Validate composer.json"
             shell: bash
             run: composer validate --no-check-publish --no-check-lock --working-dir=${{ inputs.composer-working-dir }}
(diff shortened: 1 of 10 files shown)


scheb:
I'd rather take this into a separate branch until 8.4 is actually released until all the builds can go green. Otherwise I'll have a failing build on the main branch.

Then everyone who wants to use it is free to reference that branch for time being. At least that's how I did it previously to prepare for upcoming PHP releases.

heiglandreas:
The challenge for users of the libs is that they then have to wait with testing their code until the libs are ready for PHP8.4. Which means that they won't be able to find issues with PHP8.4 while it's still in RC. Which in turn means that PHP doesn't actually know whether there are enough cases tested....

But I do have an idea that might make the build green again...

I'll address that later and perhaps we can find a way to allow testing with 8.4

heiglandreas:
The "problem" was that Psalm isn't yet ready for PHP8.4 but everything else is. As Psalm is in CI only used in the coding-standards run (which runs under PHP8.2) I removed that package completely before the default `composer install` action and then run `composer require -W --dev vimeo/psalm:^5.0` before running psalm to re-install it, now under PHP8.2.

That way it is possible to test against recent versions of PHP without having to wait for a tool that is only used with an older PHP-version to get up to speed.

The psalm-dependency is still in the composer.json so that people that want to contribute have all the tools available.

scheb:
Yea, I understand that problem and I'm also guilty on that, since I'm typically declaring compatibility with a new version, once the stable version is released. It's the classic chicken egg problem.

PHP 8.4 doesn't seem to be a big deal and I'm fine with your implementation, so let's merge it.

PR #266, opened 2025-03-21
fix: make TrustedDeviceTokenStorage conform to ResetInterface

dt-thomas-durand:
**Description**

In certain scenarios, TrustedDeviceTokenStorage might not be re-instanciated between requests. That's the case when multiple requests are handled by a single worker, like when serving the application with RoadRunner or with FrankenPHP. In that case, the trusted_device cookie leak from a request to another, and break the feature entirely. By implementing ResetInterface, the storage is now properly reseted between requests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "symfony/http-kernel": "^6.4 || ^7.0",
         "symfony/property-access": "^6.4 || ^7.0",
         "symfony/security-bundle": "^6.4 || ^7.0",
+        "symfony/service-contracts": "^2.5|^3",
         "symfony/twig-bundle": "^6.4 || ^7.0"
     },
     "require-dev": {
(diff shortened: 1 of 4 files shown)


dt-boris-castagna (approved):

scheb:
Yes, we could add this. Two things that we need to change:

- Add a unit test to ensure the values are actually reset
- Please declare `symfony/contracts` as a dependency on the package

dt-thomas-durand:
> Yes, we could add this. Two things that we need to change:
> 
> 
> 
> - Add a unit test to ensure the values are actually reset
> 
> - Please declare `symfony/contracts` as a dependency on the package
> 
> 

Thanks for your feedback.
I'll add those asap!

scheb:
Perfect, thank you very much! Released as v7.7.0.

PR #268, opened 2025-03-24
fix: added missing kernel.reset tag to scheb_two_factor.trusted_token_storage

dt-thomas-durand:
**Description**

Following #266 ; I feel shameful for not catching that mistake earlier than now…


scheb:
Released as v7.7.1

dt-thomas-durand:
> Released as v7.7.1

You're awesome!
Thanks!

PR #269, opened 2025-03-24
Update index.rst

DjLeChuck:
Add a link to our SMS provider.

It differs from the already listed `erkens/2fa-text` by using the Symfony Notifier component and offers more configuration possibility out-of-the-box.


scheb:
Sure, we can add that. Thanks!

Issue #265, opened 2025-03-20
AuthenticatorFactory priority


Tjeerd:
**Description**

Currently I am working on integrating this bundle into a project to add second factor support to the authentication layer of an API. And yes, I am also trying to keep it stateless... ;) This can be achieved fairly easy because we're only supporting the totp flow.

I've extended the main firewall config with both json_login and second_factor authenticators and am trying to trigger both authenticators in one request (which should contain username, password and second factor code). I "just" had to adjust the "flow" of the bundle a tiny bit:

* Set the `check_path` of the `two_factor` authenticator to the same `check_path` as the `json_login` authenticator
* Overwrite the `TwoFactorProviderPreparationListener` to make sure the "preparation" is run immediately (on `scheb_two_factor.authentication.attempt`) instead of `onKernelResponse` because that's too late since we're trying to authenticate immediately as well.
* Force the priority of the `TwoFactorFactory` to be lower (so it is run later) than the priority of the `JsonLoginFactory`. This way, the regular credentials are checked first, the listener is run to transform the token into a `TwoFactorToken`, and the second factor code is checked immediately after.

These adjustments work great, and I got a stateless login flow with second factor support!

First two steps are easy to do, but unfortunately the last step isn't. I did not find a way of overwriting the priority of the authenticator (without adjusting the code of the bundle ofc).

Currently the priority is `0`. The json_login priority is `-40`. I am thinking that the two factor authenticator should (almost?) always come last if there are multiple authenticators. Do you agree, and if so, would you accept a PR to lower the priority of the authenticator to, say, `-50`?

scheb:
I can absolutely see the point of this change. A single-step authentication process for stateless firewalls would be pretty nice.

I tried to make the priority a configurable option, but unfortunately the security extension is loaded, i.e. the priority is read, before any bundle configuration has been initialized. That would have been the backwards-compatible way to introduced the change. You can see my failed try in the [authenticator-priority](https://github.com/scheb/2fa/compare/7.x...authenticator-priority) branch.

Except we find another way how to introduce this in a backwards-compatible way, I wouldn't want to change the priority outside of a major release, because this could be a breaking change for people's applications, when the order of their authentication providers suddenly changes. The consequences are hard to estimate. It could be totally no problem at all. Could also be, that it causes issues with authentication in certain setups. Hard to tell 🤷. Considering the reach of the bundle, I wouldn't want to make this change and "hope for the best".

Tjeerd:
[...]

PR #273, opened 2025-04-07
Allow preparation of providers to be optional

Tjeerd:
**Description**

To pave the way for stateless 2nd factor authentication the bundle can be made aware of providers that do no need to be prepared.

These providers can be used for stateless auth because there is no need to store the state of the preparation.

This MR skips calling the preparation, storing the state of the preparation and checking this state if the provider does not need to be prepared. It does this by adding a bool method `needsPreparation` to the `ProviderInterface` and checking this method where appropriate.

Unfortunately this is a backwards incompatible change and would require other Provider implementations to add the method.

Also see #265

### DIFF
--- a/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php
+++ b/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php
@@ -8,6 +8,8 @@ use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenFactoryInt
 use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\AuthenticationContextInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Exception\UnknownTwoFactorProviderException;
+use function array_walk;
+use function count;
 
 /**
  * @final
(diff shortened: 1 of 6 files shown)


scheb:
I have been thinking about this myself when I started the 8.x todo list :)

As you mentioned, this will be a BC break, so I'd consider this for the version 8 major release.

Though my approach would be a bit different. I'd set all the authentication providers that do _not_ need preparation to "prepared" when the security token is created (in `Scheb\TwoFactorBundle\Security\TwoFactor\Provider\TwoFactorProviderInitiator::beginTwoFactorAuthentication`). Then all the other code can just stay as it is. Should overall have less complexity, because you don't need to spread "does this provider need preparation" logic at multiple places.

Tjeerd:
> Though my approach would be a bit different. I'd set all the authentication providers that do _not_ need preparation to "prepared" when the security token is created (in `Scheb\TwoFactorBundle\Security\TwoFactor\Provider\TwoFactorProviderInitiator::beginTwoFactorAuthentication`). Then all the other code can just stay as it is. Should overall have less complexity, because you don't need to spread "does this provider need preparation" logic at multiple places.

I considered the same approach :) But choose to do it this way, because it feels a bit "cleaner" and represents the actual behavior instead of "faking" the preparation state. But I agree as well: it would probably require less changes and therefore probably introduce less bugs 😅

scheb:
Yea, I feel the same issue with the naming/semantics. Would be better if it's called "is ready" or something like that, instead of calling it "prepared", when there was actually no preparation work done. Though, you could argue something that doesn't need preparation could be considered "automatically prepared" 🤷

We could rename things (and introduce more BC breaks), but that's probably not worth bothering too much with it. I'd definitely prefer a less complex implementation over the cleaner naming.

scheb:
Looks much cleaner, definitely that approach 👍 

Things that I'd change before merging:
- I'd evaluate the "needs preparation" when the active 2fa providers are collected, so we don't need to iterate the providers twice. The return value of the private `getActiveTwoFactorProviders` method would need to change for this. Or the code is reorganized as a whole. I believe the `getActiveTwoFactorProviders` is only there for historic reasons, as this used to be a public method.
- Unit test coverage :)

What would you suggest how to move forward? I can't merge this into `7.x` and a version 8 is not going to appear that soon. I could create a `8.x` development branch and we could merge it there. What do you think?

Tjeerd:
Agreed, I'll add some tests and I've updated the loop to be a little more efficient.

I noticed this MR now also include the prio change of the listener. To be honest this was not necessarily the intention, but maybe we can use this MR to "Add support for stateless second factor authentication". In this case we should probably also add some documentation about how to setup such authentication. :)

Tjeerd:
About that branching, I'll leave that up to you :)

I've integrated my MR branch into an actual project to test it "in production". If this branch/MR lands somewhere in this repo, I'll switch my project to this branch as well.

src/bundle/DependencyInjection/Factory/Security/TwoFactorFactory.php
```diff
@@ -169,6 +169,6 @@ public function getKey(): string
 
     public function getPriority(): int
     {
-        return 0;
+        return -50;
```
scheb: This piece needs to be removed. Authenticator priority has already been changed on the 8.x branch.

src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInterface.php
```diff
@@ -13,6 +13,11 @@ interface TwoFactorProviderInterface
      */
     public function beginAuthentication(AuthenticationContextInterface $context): bool;
 
+    /**
+     * Determine whether this Provider needs to be prepared (if the prepareAuthentication method needs to be called).
+     */
+    public function needsPreparation(): bool;
+
     /**
      * Do all steps necessary to prepare authentication, e.g. generate & send a code.
```
scheb: To make the new behavior more clear:

```suggestion
     * Do all steps necessary to prepare authentication, e.g. generate & send a code. Will only be called when needsPreparation() returns true.
```

src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php
```diff
@@ -21,46 +23,42 @@ public function __construct(
...
                 continue;
             }
 
             $activeTwoFactorProviders[] = $providerName;
+
+            if (!$provider->needsPreparation()) {
+                continue;
+            }
+
+            $statelessProviders[] = $providerName;
```
scheb: I believe the logic is wrong in this place and that's why the integration test is failing. Actually it should be the other way round.

It's easy to get confused here, because the "return early" approach makes it harder to understand which values go into the array. Would rather implement it like this, to make the condition for the `$statelessProviders` array more clear:

```suggestion
            if (!$provider->needsPreparation()) {
                $statelessProviders[] = $providerName;
            }
```

scheb (changes requested):
I've updated the MR to go into the `8.x` branch. The `8.x` is already containing the change for authenticator priority. So please do a rebase on your side.

src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php
Tjeerd: Check, my bad 👍

scheb:
Perfect 👍. Integration tests have also gone green.

Unit tests need to be added, then we're ready to merge.

scheb:
Perfect, thank you! 🙇

PR #274, opened 2025-04-07
Allow preparation of providers to be optional

Tjeerd:
See #273, a much simpler approach :)


scheb:
Could you force-push that on the already existing MR, so we don't lose context?

PR #275, opened 2025-04-27
Update totp.rst

gturri:
The constructor of TotpConfiguration expects the first parameter $secret to be a string. So if it is null, it will generate an error at runtime. OTHO the method TwoFactorInterface.getTotpAuthenticationUsername() is expected to return a ?TotpConfigurationInterface, so it's ok if we return null in this situation

**Description**

### DIFF
--- a/doc/providers/totp.rst
+++ b/doc/providers/totp.rst
@@ -93,7 +93,7 @@ and the period of the temporary codes.
            public function getTotpAuthenticationConfiguration(): ?TotpConfigurationInterface
            {
                // You could persist the other configuration options in the user entity to make it individual per user.
-               return new TotpConfiguration($this->totpSecret, TotpConfiguration::ALGORITHM_SHA1, 20, 8);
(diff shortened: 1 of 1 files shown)


doc/providers/totp.rst
```diff
@@ -93,7 +93,7 @@ and the period of the temporary codes.
            public function getTotpAuthenticationConfiguration(): ?TotpConfigurationInterface
            {
                // You could persist the other configuration options in the user entity to make it individual per user.
-               return new TotpConfiguration($this->totpSecret, TotpConfiguration::ALGORITHM_SHA1, 20, 8);
+               return $this->isTotpAuthenticationEnabled() ? new TotpConfiguration($this->totpSecret, TotpConfiguration::ALGORITHM_SHA1, 20, 8) : null;
```
scheb: I see the point. Though if the secret being null is the issue, better check for `$this->totpSecret` instead of $this->isTotpAuthenticationEnabled(). That makes is more explicit what is going on.

```suggestion
               return null !== $this->totpSecret ? new TotpConfiguration($this->totpSecret, TotpConfiguration::ALGORITHM_SHA1, 20, 8) : null;
```

scheb:
Thanks! Merged.

Issue #277, opened 2025-05-06
Make sure to only allow single usage of a 2FA code


heiglandreas:
Right now it is possible to use a 2FA code several times as long as the code itself isn't expired. 

According to [NIST](https://pages.nist.gov/800-63-3/sp800-63b.html#5152-multi-factor-otp-verifiers) though that should not be possible. 

> Time-based OTPs [RFC 6238] SHALL have a defined lifetime that is determined by the expected clock drift — in either direction — of the authenticator over its lifetime, plus allowance for network delay and user entry of the OTP. **In order to provide replay resistance as described in Section 5.2.8, verifiers SHALL accept a given time-based OTP only once during the validity period**. In the event a claimant’s authentication is denied due to duplicate use of an OTP, verifiers MAY warn the claimant in case an attacker has been able to authenticate in advance. Verifiers MAY also warn a subscriber in an existing session of the attempted duplicate use of an OTP. (emphasis mine)

If I see that correctly (please correct me!) there is currently no way of getting this behaviour out of the box from this package. 

Would it not be an idea to provide a simple implementation of this behaviour using an existing caching method to implement this? I was thinking about providing a `Psr\Cache\CacheItemPoolInterface` or a `Symfony\Contracts\Cache\CacheInterface` in the package-config and when that is available to use that for ensuring single use of the code.

I will have to implement it either way, would it be of interest to do that in a way that it can be added to the package?

scheb:
This was discussed quite recently in #276.

Short version: There is currently no out-of-the-box mechanic to keep track of used authentication codes and prevent reuse. For custom implementations, it could be achieved by listening to the newly introduced "check" events and throwing an exception, when a reused code is detected.

I'd be open to make this part of the bundle, though I'd be reluctant to introduce a complicate storage mechanism or new dependency to the bundle. Relying on `symfony/cache` infrastructure seems like a good approach to me, since it's pulled in as a dependency anyways.

For a proper implementation I'd likely add another listener like `CheckTwoFactorCodeListener`, which checks codes for reuse and throws an exception of `ReusedTwoFactorCodeException extends BadCredentialsException`. That should make the 2fa authentication fail and make the respective error message show up in the UI.

heiglandreas:
[...]

PR #279, opened 2025-05-07
Handle multiple usages of an MFA-code

heiglandreas:
This PR introduces the possibility to handle reuse of MFA-codes as discussed in #277

I also made the default event-handler for the `TwoFactorCodeReusedEvent` configurable to allow easier disabling. 

The tests were passing, PHPCodesniffer was happy as well, the docs are updated and I hope I didn't butcher the config too much.

Any feedback is much appreciated!


scheb:
Haven't gone through the code yet, since you're obviously still working on it ;)

One thing that I can already tell for sure, we'd need a `code_reuse_protection: true/false` configuration option, to switch that feature on/off. Should be default off, so the bundle's behavior doesn't change when people update to the new version. Cache/handle can then be preconfigured with their default values.

heiglandreas:
Yeah. I am right now debugging a tiny detail (doesn't work at all 😂)

But the idea was that we don't need a `code_reuse_protection` switch. As long as no Cache-Provider is given, no check is done so nothing changes.

Otherwise we'd have 3 config-values: the flag, the cache-provider and the default-handler. And as one needs to provide the cache-provider anyhow. 

I'd rather not assume the `cache.app` as default. That might not be set and therefore doesn't work out of the box. So people might enable the protection but it doesn't work at all. I'd rather avoid that by explicitly requiring users to provide a cache-implementation.

heiglandreas:
All issues fixed from my side.

Works in a real life scenario

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
```diff
@@ -0,0 +1,65 @@
...
+    public function checkForCodeReuse(TwoFactorCodeCheckEvent $event): void
+    {
+        if (null === $this->cache) {
+            return;
+        }
+
+        $cacheItem = null;
+        assert($cacheItem instanceof CacheItemInterface || null === $cacheItem);
+        $cacheKey = 'scheb_two_factor_code_reuse.'
+            .sha1($event->getUser()->getUserIdentifier().'.'.$event->getCode());
```
heiglandreas: As the useridentifier can contain characters that are not allowed I decided to create a hash from the user-identifier and the code. The result is the same, a unique key for a combination of username and 2FA-code. 

The added benefit is, that neither value will be exposed in the cache so from a security point of view this is better than using the username and the token.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
```diff
@@ -0,0 +1,65 @@
...
+
+        if ($this->cache instanceof CacheItemPoolInterface) {
+            $cacheItem = $this->cache->getItem($cacheKey);
+            $cacheItem->expiresAfter(60);
```
heiglandreas: The 60 seconds are an arbitrary number. But as most tokens are configured to invalidate after 30 seconds that should be enough. If need be we could also make this configurable but the less moving parts the less confusing it is for the users.

doc/events.rst
```diff
@@ -206,3 +206,23 @@ Constant: ``Scheb\TwoFactorBundle\Security\TwoFactor\Event\TotpCodeEvents::INVAL
...
+
+Is dispatched before the TOTP authentication provider is used to check the code for plausibility
+
+``TwoFactorCodeReusedEvent``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Event class: ``\Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorCodeReusedEvent``
+
+Is dispatched when the code has been used already within the last 30 seconds.
+This requires a caching backend to be available
```
scheb: For consistency with the other events, these two events should have an event name as well. Suggestion:
- `scheb_two_factor.authentication.check`
- `scheb_two_factor.authentication.code_reused`

Then, I'd move the whole block up after `Authentication Events` and relabel it `Code Check Events`, rather than `TOTP Check Events`, since these are generic events that are dispatched for all kinds of authentication providers. Not just the TOTP one.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
scheb: Good choice. The security aspect is also a good point.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
scheb: 60 is certainly a fair choice, but I'd rather make this a configuration option. Maintaining this bundle has taught me, you never know what kind of "weird" things people do in their applications. There will be the person, who has a 120 seconds TOTP period, who will complain 🙈. So providing a configuration option with default 60 is the safe path forward.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function checkForCodeReuse(TwoFactorCodeCheckEvent $event): void
+    {
+        if (null === $this->cache) {
+            return;
+        }
+
+        $cacheItem = null;
+        assert($cacheItem instanceof CacheItemInterface || null === $cacheItem);
```
scheb: What's the assert for? `$cacheItem` is always set `null` right before, so it doesn't do much.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
```diff
@@ -0,0 +1,65 @@
...
+        $cacheKey = 'scheb_two_factor_code_reuse.'
+            .sha1($event->getUser()->getUserIdentifier().'.'.$event->getCode());
+
+        if ($this->cache instanceof CacheItemPoolInterface) {
```
scheb: I'm wondering what we should do if the cache object is something else. I believe it would be good to have at least a warning, probably better an error, logged so it doesn't fail silently.

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+    protected function expectDoNothing(): void
+    {
+        $this->eventDispatcher
+            ->expects($this->never())
+            ->method($this->anything());
+    }
+
+    #[Test]
+    public function nothingHappensWithoutCacheProvider(): void
```
scheb: Mentioning it here for all the test methods: Test methods should be named after the schema `[method]_[scenario]_[expectedResult]`, to be consistent with the naming convention of the tests.

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
```diff
@@ -0,0 +1,107 @@
...
+use Scheb\TwoFactorBundle\Tests\TestCase;
+use stdClass;
+use Symfony\Component\Cache\CacheItem;
+use Symfony\Component\Security\Core\User\UserInterface;
+use function sha1;
+
+/**
+ * @property CheckTwoFactorCodeListener $listener
+ */
+class CheckTwoFactorCodeReuseListenerTest extends TestCase
```
scheb, line 27: There should be at least one test to make sure the cache is written, meaning the `save()` method is called with a CacheItem that has the expected values.

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $cacheProvider = $this->createMock(CacheItemPoolInterface::class);
+        $cacheItem = new CacheItem();
+        $cacheProvider->expects($this->once())
+            ->method('getItem')->willReturnCallback(function (string $key) use ($cacheItem) {
+                $this->assertEquals('scheb_two_factor_code_reuse.'.sha1('123.123456'), $key);
+                $cacheItem->set(true);
+                $cacheItem->expiresAfter(60);
+
+                return $cacheItem;
+            });
```
scheb: For ` 123` and `123456` descriptive constants should be used to make clear what these values represent. `123` would also need to be extracted into a `USER_IDENTIFIER` constant.

I'm also wondering, if the callback is really necessary. For testing the "no hit" behavior, it might even be sufficient to go with the same approach as below and just return a mocked CacheItem = `$cacheItem->expects($this->once())->method('isHit')->willReturn(false);`.

tests/Security/Http/EventListener/ThrowExceptionOnTwoFactorReuseListenerTest.php
```diff
@@ -0,0 +1,40 @@
...
+ * @property CheckTwoFactorCodeListener $listener
+ */
+class ThrowExceptionOnTwoFactorReuseListenerTest extends TestCase
+{
+    #[Test]
+    public function throwExceptionWhenUsed(): void
+    {
+        $listener = new ThrowExceptionOnTwoFactorCodeReuseListener();
+
+        $this->expectException(ReusedTwoFactorCodeException::class);
```
scheb: Better to test the whole event here, to make sure the properties (i.e. code) are set, like you did for the `TwoFactorCodeReusedEvent` in `CheckTwoFactorCodeReuseListenerTest`.

tests/Security/Http/EventListener/ThrowExceptionOnTwoFactorReuseListenerTest.php
```diff
@@ -0,0 +1,40 @@
...
+        ));
+    }
+
+    #[Test]
+    public function EventIsHandledWithCorrectPriority(): void
+    {
+        $this->assertSame([
+            TwoFactorCodeReusedEvent::class => ['handle', -256],
+        ], ThrowExceptionOnTwoFactorCodeReuseListener::getSubscribedEvents());
+    }
```
scheb: I wouldn't require to test those `getSubscribedEvents` methods, because they're just glue code without logic. You don't have to remove it, but also no need to keep it, since the test doesn't provide much value ;)

src/bundle/Security/Authentication/Exception/ReusedTwoFactorCodeException.php
```diff
@@ -0,0 +1,21 @@
...
+
+use Symfony\Component\Security\Core\Exception\BadCredentialsException;
+
+/**
+ * @final
+ */
+class ReusedTwoFactorCodeException extends BadCredentialsException
+{
+    public const MESSAGE = 'Reused two-factor authentication code.';
+    private const MESSAGE_KEY = 'code_reused';
```
scheb, line 15: The translation string should be added in `SchebTwoFactorBundle.en.yml`. And if you can add extra translations, would be much appreciated. I could provide a German translation, and I guess so could you ;)

scheb (commented):
Sorry for the late response! I got distracted and forgot about the PR. My bad!

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
heiglandreas: I think that was a change by PHP-cs-fixer or PHPCS 🙈 

Will remove it....

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
heiglandreas: Triggering a warning is probably indeed the best idea.

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
heiglandreas: Yeah. This tests in essence two different things:

* The `getItem` is called with the expected data 
* On the first invocation with a 2FA-code nothing happens as expected.

For that I need to make sure that the passed `$key` is actually the expected one. 

But I can do that similar to the `codeReusedEventIsDispatchedWithCacheHit` method using `with` and `return`.

tests/Security/Http/EventListener/ThrowExceptionOnTwoFactorReuseListenerTest.php
heiglandreas: That was just to get to a decent LOC-coverage of over 50% 😁

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
heiglandreas: Logging an error now.

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
heiglandreas: Test is added

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
heiglandreas: Simplified this test

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
heiglandreas: Implemented

doc/events.rst
heiglandreas: Good point!

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
```diff
@@ -0,0 +1,151 @@
...
+        $cacheProvider->expects($this->once())
+            ->method('save')
+            ->willReturnCallback(function (CacheItem $item) {
+                $this->assertFalse($item->isHit());
+                $this->assertSame(true, $item->get());
+                $sweetsThief = Closure::bind(function (CacheItem $item) {
+                    return $item->expiry;
+                }, null, CacheItem::class);
+                $this->assertGreaterThanOrEqual(time() + 20, $sweetsThief($item));
+                $this->assertLessThanOrEqual(time() + 21, $sweetsThief($item));
```
heiglandreas: I didn't find an easier way to handle this as the correct lifetime should be checked here. And getting that from the CacheItem is only possible by either creating an own implementation of the CacheItemInterface or using the existing one. Using the existing one in combination with this abomination was less code, so I went that route. 

Can change that though if you think it'd be better for future maintenance.

tests/Security/Http/EventListener/ThrowExceptionOnTwoFactorReuseListenerTest.php
```diff
@@ -0,0 +1,46 @@
...
+        try {
+            $listener->handle(new TwoFactorCodeReusedEvent(
+                $this->createMock(UserInterface::class),
+                self::MFA_CODE,
+            ));
+        } catch (ReusedTwoFactorCodeException $exception) {
+            $this->assertSame(0, $exception->getCode());
+            $this->assertSame('code_reused', $exception->getMessageKey());
+            $this->assertSame('', $exception->getMessage());
+        }
```
heiglandreas, line 37: I went this approach instead of the `expectedException*` methods as the `getMessageKey` would otherwise not be testable.

doc/events.rst
```diff
@@ -66,6 +66,30 @@ Event class: ``Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorAuthentic
...
+
+``scheb_two_factor.authentication.code_reused``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Constant: ``Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorAuthenticationEvents::CODE_REUSED``
+
+Event class: ``\Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorCodeReusedEvent``
+
+Is dispatched when the code has been used already within the last 30 seconds.
+This requires a caching backend to be available
```
scheb:
```suggestion
Is dispatched when the code has already been used by the same user within the configured time frame.
This requires a caching backend to be available
```

src/bundle/DependencyInjection/Configuration.php
```diff
@@ -58,6 +58,9 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->scalarNode('two_factor_token_factory')->defaultValue('scheb_two_factor.default_token_factory')->end()
                 ->scalarNode('two_factor_provider_decider')->defaultValue('scheb_two_factor.default_provider_decider')->end()
                 ->scalarNode('two_factor_condition')->defaultNull()->end()
+                ->scalarNode('code_reuse_cache')->defaultNull()->end()
+                ->integerNode('code_reuse_cache_duration')->defaultValue(60)->min(30)->end()
```
scheb: Is there a technical reasoning for the 30 minimum? 🤔 If not, I wouldn't restrict that, as it seems to be a bit over-restrictive to me.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
```diff
@@ -0,0 +1,70 @@
...
+            $cacheItem = $this->cache->getItem($cacheKey);
+            $cacheItem->expiresAfter($this->cacheDuration);
+            $cacheItem->set(true);
+            $this->cache->save($cacheItem);
+        }
+
+        if (!$cacheItem instanceof CacheItemInterface) {
```
scheb: I believe this should be the `else` branch, executed whenever `$this->cache` is not a supported interface.

```suggestion
        } else {
```

src/bundle/DependencyInjection/SchebTwoFactorExtension.php
```diff
@@ -234,6 +235,28 @@ private function configureTotpAuthenticationProvider(ContainerBuilder $container
...
+     */
+    private function configureCodeReuseCache(ContainerBuilder $container, array $config): void
+    {
+        if (($config['code_reuse_cache'] ?? null) === null) {
+            $config['code_reuse_cache'] = '';
+        }
+
+        if (($config['code_reuse_cache_duration'] ?? null) !== null) {
+            $container->setParameter('scheb_two_factor.code_reuse_cache_duration', (int) $config['code_reuse_cache_duration']);
+        }
```
scheb: To my understanding `$config['code_reuse_cache_duration']` will always be an integer. When the config value isn't set, the default integer value will be provided. So I believe the parameter can always be set without any extra checks.

```suggestion
        $container->setParameter('scheb_two_factor.code_reuse_cache_duration', $config['code_reuse_cache_duration']);
```

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            return;
+        }
+
+        if (!$cacheItem->isHit()) {
+            return;
+        }
+
+        $this->eventDispatcher->dispatch(new TwoFactorCodeReusedEvent($event->getUser(), $event->getCode()));
```
scheb: The early return makes the logic harder to understand. Would actually put the dispatch inside the if, so make it clear what's going on: When there is a cache hit -> dispatch event. 

```suggestion
        if ($cacheItem->isHit()) {
            $this->eventDispatcher->dispatch(new TwoFactorCodeReusedEvent($event->getUser(), $event->getCode()));
        }
```

src/bundle/Security/TwoFactor/Event/TwoFactorCodeCheckEvent.php
```diff
@@ -0,0 +1,29 @@
...
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * @final
+ */
+class TwoFactorCodeCheckEvent extends Event
```
scheb: I'm not sure about `TwoFactorCodeCheckEvent` and `TwoFactorCodeReusedEvent`. The already existing `TwoFactorCodeEvent` has the same structure and the same intention ("something is happening with a 2fa code"), so I'd probably use this class, instead of introduced two new classes.

What's your thoughts about this?

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
```diff
@@ -0,0 +1,151 @@
...
+            ->method('dispatch')
+            ->with(new TwoFactorCodeReusedEvent($this->user, self::MFA_CODE));
+
+        $listener = new CheckTwoFactorCodeReuseListener($this->eventDispatcher, $cacheProvider, 60, null);
+
+        $listener->checkForCodeReuse(new TwoFactorCodeCheckEvent($this->user, self::MFA_CODE));
+    }
+
+    #[Test]
+    public function checkForCodeREuse_validCacheProviderNoCachewHit_CacheItemIsSaved(): void
```
scheb: Typos ;)
```suggestion
    public function checkForCodeReuse_validCacheProviderNoCacheHit_cacheItemIsSaved(): void
```

tests/Security/Http/EventListener/CheckTwoFactorCodeReuseListenerTest.php
scheb: We might be better off, injecting a mocked CacheItem as a "spy" (untested code written in GitHub UI 😅), though this approach has the disadvantage, not making sure the values are updated _before_ the save is executed. No perfect solution here. Would also be fine keeping yours. Thoughts?

Assertions for the event dispatching I'd completely remove from this test, as it's already covered in its own dedicated test case.

```suggestion
    #[Test]
    public function checkForCodeReuse_validCacheProviderNoCachewHit_cacheItemIsSaved(): void
    {
        $cacheItem = $this->createMock(CacheItem::class);
        //If you want to be fancy, run the test for both isHit true and false, so make sure in both cases, the expiration time is updated
        // $cacheItem
        //     ->expects($this->once())
        //     ->method('isHit')
        //     ->with(true/false);
        
        $cacheItem
            ->expects($this->once())
            ->method('set')
            ->with(true);
        $cacheItem
            ->expects($this->once())
            ->method('expiresAfter')
            ->with(20);
        
        $cacheProvider->expects($this->once())
            ->method('getItem')
            ->with('scheb_two_factor_code_reuse.3468ccbd044e2fdfb0769b68b5c85d7660c6c12c')
            ->willReturn($cacheItem);
        
        $cacheProvider = $this->createMock(CacheItemPoolInterface::class);
        $cacheProvider->expects($this->once())
            ->method('save')
            ->with($cacheItem);
            
        // Event behavior already covered in the other two tests, so not testing here

        $listener->checkForCodeReuse(new TwoFactorCodeCheckEvent($this->user, self::MFA_CODE));
    }
```

scheb (commented):
Overall, almost ready. Have some details left, that we should discuss.

doc/events.rst
heiglandreas: I deliberately left the "same user" out 😁. But I don't really have a strong opinion, so absolutely fine with that

src/bundle/DependencyInjection/Configuration.php
heiglandreas: Nope! Apart from making sure people don't shoot themselves right in the middle of the foot...

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
heiglandreas: That is a leftover from when I checked for 2 different interfaces and will be changed!

src/bundle/Security/TwoFactor/Event/TwoFactorCodeCheckEvent.php
heiglandreas: I was thinking about that and decided against reusing the `TwoFactorCodeEvent`.

While the three events share the same code they do not share the same intention. The same code is for me in essence just an implementation detail. 

Besides that I found it easier to have dedicated classes for dedicated events as that makes the whole coding IMO easier. The EventHander or -Subscriber only checks the class and there is no need to also check the event-name. And when dispatching an event I, well, dispatch an event. No need to also use an event-name. Otherwise I dispatch a DTO along with an event-name

But that is definitely not a hill I'll die on!

It's your bundle and as you used a different approach so far I have no issue in changing that.

So I'd use the names to distinguish and the `TwoFactorCodeEvent` as DTO.

src/bundle/DependencyInjection/Configuration.php
scheb: Ok, then let's remove it. Configuration reference docs also need to be updated, as it mentions the minimum.

src/bundle/Security/TwoFactor/Event/TwoFactorCodeCheckEvent.php
scheb: Let's use `TwoFactorCodeEvent` instead of `TwoFactorCodeCheckEvent`, because it is the "code checking" event class in the bundle.

We keep `TwoFactorCodeReusedEvent`, because it's a whole different signal ("something is wrong with the code", instead of "I'm checking the code").

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
heiglandreas: I now know where that came from 😂 

PHP_COdeSniffer complaints about this and phpcbf auto"fixes" it...

```
 56 | ERROR | [x] Use early exit to reduce code nesting.
```

Shall I add a comment to make sure that it stays that way?

src/bundle/Security/TwoFactor/Event/TwoFactorCodeCheckEvent.php
heiglandreas: But to stay consistent I'D then use the name in all cases and "just" add the object as DTO. Also for the `TwoFactorCodeReuseEvent`...

scheb: Yes please, dispatch with event name everywhere. To do that consistently with the rest of the bundle.

src/bundle/Security/Http/EventListener/CheckTwoFactorCodeReuseListener.php
scheb: Well, in that case I'd add a comment to suppress the phpcs error, because I'd value the clarity over "less nesting".

src/bundle/Security/TwoFactor/Event/TwoFactorCodeCheckEvent.php
heiglandreas: Another reason for using separate EventObjects: 

```
Psr\EventDispatcher\EventDispatcherInterface::dispatch(object $event)
```

No name... That's only available in the `Symfony\Contracts\EventDispatcher\EventDispatcherInterface`

scheb: I know, this has changed when the PSR event dispatcher interface has been introduced to the Symfony ecosystem. The "traditional" Symfony way is with event names + event object, with the event name identifying the type of event and the class just being some DTO. This is still supported by Symfony API, see `Symfony\Contracts\EventDispatcher\EventDispatcherInterface`. 

👉 Now that I write this, we should change the reference to `Symfony\Contracts\EventDispatcher\EventDispatcherInterface` instead of `Psr\EventDispatcher\EventDispatcherInterface`.

When time comes, I'll eventually get rid of all the event names and switch to the PSR interface. But for now, I'd want to keep it consistent with event name + event class everywhere in the bundle.

heiglandreas: Already switched 😁 

And yes! Consistency is king!

heiglandreas:
Ping @scheb: Anything still missing?

scheb:
I wasn't sure if you're already done. Will have a look now.

doc/configuration.rst
```diff
@@ -105,6 +105,24 @@ Bundle Configuration
...
+       # Remove the key or set the value to an empty string to disable this check
+       code_reuse_cache: acme.custom_code_reuse_cache
+
+       # By default the time to cache a 2FA-code is 60 seconds. After that the 2FA code should be
+       # invalidated by the 2FA provider anyhow. Should your 2FA provider allow valid codes for
+       # more than that, you should increase this amount!
+       code_reuse_cache_duration: 60
+
+       # If you need to handle reused codes differently you can create your own EventListener for the
+       # Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorCodeReusedEvent
```
scheb: Should mention the event name here, since that is what you need to subscribe to:

```suggestion
       # scheb_two_factor.authentication.code_reused
```

src/bundle/Resources/config/security.php
```diff
@@ -67,7 +69,21 @@
...
+                service('event_dispatcher'),
+                service('scheb_two_factor.code_reuse_cache')->nullOnInvalid(),
+                '%scheb_two_factor.code_reuse_cache_duration%',
+                service('logger')->nullOnInvalid(),
+            ])
+
+        ->set('scheb_two_factor.security.listener.throw_exception_on_two_factor_code_reuse', ThrowExceptionOnTwoFactorCodeReuseListener::class)
+            ->tag('kernel.event_subscriber')
+
+        ->alias('scheb_two_factor.security.listener.default_code_reuse_listener', 'scheb_two_factor.security.listener.throw_exception_on_two_factor_code_reuse')
```
scheb: The way this is implemented, I believe, setting a custom event handler wouldn't disable the default `ThrowExceptionOnTwoFactorCodeReuseListener` handler. This is because `ThrowExceptionOnTwoFactorCodeReuseListener` is registered as its own service, which will remain active in the dependency injection container and therefore be executed, even when a custom event handler was set on the alias.

For this to work as intended, when setting a custom event handler, the `scheb_two_factor.security.listener.throw_exception_on_two_factor_code_reuse` definition would actually need to be removed.

The alias is effectively purely informational and doesn't have any effect on the application's behavior.

So I believe we should:
- Remove the alias, since it's not needed and adds clutter
- Whenever a custom event listener is configured, remove the `scheb_two_factor.security.listener.throw_exception_on_two_factor_code_reuse` definition or at least the event subscriber tag to disable the listener

scheb (commented):
Rest looks good. Smart solution with the fake CacheItemInterface 👍

src/bundle/Resources/config/security.php
heiglandreas: Thanks for noticing. This should now be handled the "right" may. I'm completely removing the default event-listener from the container as untagging seems not to be possible. But as people want to use a different listener that's totally fine in the end...

doc/configuration.rst
heiglandreas: Done. I used the Class-Constant as I'D rather have people use that instead of the literal string but it should result in the same thing.

doc/configuration.rst
```diff
@@ -105,6 +105,24 @@ Bundle Configuration
...
+       # By default the time to cache a 2FA-code is 60 seconds. After that the 2FA code should be
+       # invalidated by the 2FA provider anyhow. Should your 2FA provider allow valid codes for
+       # more than that, you should increase this amount!
+       code_reuse_cache_duration: 60
+
+       # If you need to handle reused codes differently you can create your own EventListener for the
+       # Scheb\TwoFactorBundle\Security\TwoFactor\Event\TwoFactorAuthenticationEvents::CODE_REUSED event
+       # The default handler throws a Scheb\TwoFactorBundle\Security\Authentication\Exception\ReusedTwoFactorCodeException
+       # If that is not what you want, you can provide a different default handler here.
+       code_reuse_default_handler: Scheb\TwoFactorBundle\Security\Http\EventListener\ThrowExceptionOnTwoFactorCodeReuseListener
```
scheb: The code in `SchebTwoFactorExtension.php` is expecting the service name here, so this should be `scheb_two_factor.security.listener.throw_exception_on_two_factor_code_reuse`

```suggestion
       code_reuse_default_handler: scheb_two_factor.security.listener.throw_exception_on_two_factor_code_reuse
```

scheb (commented):
One last change 😅

scheb:
v7.10.0 released :)

heiglandreas:
🎉

Thank you so much for bearing with me and guiding me through the process!

PR #280, opened 2025-05-11
provide default values for totp configuration

gturri:
**Description**

The doc (at doc/providers/totp.rst) says that there are some default value. This commit ensure those are the values actually used if the user does not provide those.
It also update a snippet in that doc to ensure its clear for readers what each value stands for.

### DIFF
--- a/doc/providers/totp.rst
+++ b/doc/providers/totp.rst
@@ -94,7 +94,9 @@ and the period of the temporary codes.
            public function getTotpAuthenticationConfiguration(): ?TotpConfigurationInterface
            {
                // You could persist the other configuration options in the user entity to make it individual per user.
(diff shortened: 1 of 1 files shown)


scheb:
I'm fine with the documentation change, but the `TotpConfiguration` should be left as is. People should explicitly set these values, so they know what they get, rather than relying on somewhat "hidden" default values.

I'll update the line in the documentation (-> https://github.com/scheb/2fa/commit/7ef0ee7395bc7fd4222ae2b2f2bac2eb112ed8b3), which is written in a misleading way. It's not meant to be "technical defaults", which are provided with the bundle. The intention was rather to let people know that this is the most common configuration for TOTP out there.

gturri:
thanks for this feedback.
Here is an update of this PR, with just the doc updated

scheb:
Awesome, thanks!

Issue #283, opened 2025-06-13
`TwoFactorInProgressVoter::vote()` is deprecated in Symfony 7.3


jkgroupe:
**Bundle version**: 7.10.0
**Symfony version**: 7.3
**PHP version**: 8.2

**Description**
The function `vote(TokenInterface $token, mixed $subject, array $attributes)` in `Scheb\TwoFactorBundle\Security\Authorization\Voter\TwoFactorInProgressVoter` is deprecated in Symfony 7.3 and must be replaced by `vote(TokenInterface $token, mixed $subject, array $attributes, ?Vote $vote = null)`

![Image](https://github.com/user-attachments/assets/d5f00759-566c-4a67-947f-85d698254fcd)

PS : sorry, I choosed the wrong label

scheb:
Addressed in v7.10.1

Issue #282, opened 2025-06-04
`TokenInterface::eraseCredentials()` is deprecated in symfony 7.3


W0rma:
[The 2fa-bundle](https://github.com/scheb/2fa/blob/7.x/src/bundle/Security/Authentication/Token/TwoFactorToken.php) implements `TokenInterface::eraseCredentials()` which is deprecated since https://github.com/symfony/symfony/pull/59682

This leads to the following deprecation warning if symfony 7.3 is installed:

> Since symfony/security-http 7.3: Implementing "Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorToken::eraseCredentials()" is deprecated since Symfony 7.3; add the #[\Deprecated] attribute on the method to signal its either empty or that you moved the logic elsewhere, typically to the "__serialize()" method.

scheb:
Addressed in v7.10.1

gnutix:
I'm getting this deprecation even though I'm on Symfony 6.4 (using 7.11.0) : 

> PHP Deprecated:  Method Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorToken::eraseCredentials() is deprecated in vendor/symfony/security-http/Authentication/AuthenticatorManager.php on line 204

Should I open a separate issue?

scheb:
It is likely cause by this change https://github.com/scheb/2fa/pull/284, which is supposed to address the deprecation notice on Symfony 7.3+. But if you're saying, you see a deprecation notice with Symfony 6.4, then this is then likely causing deprecation notices into the other direction (Symfony 7.2 and lower) now. Obviously, you can't have both addressed at the same time 🤷.

So in the end, my recommendation to you would be to adjust your logging settings, so that deprecation notices (which are - notices - not errors), are not bothering you.

PR #284, opened 2025-06-23
Fix Symfony 7.4 deprecation

jkgroupe:
**Description**
Fix Symfony 7.4 deprecation from issues #283 and #282

### DIFF
--- a/src/bundle/Security/Authentication/Token/TwoFactorToken.php
+++ b/src/bundle/Security/Authentication/Token/TwoFactorToken.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Scheb\TwoFactorBundle\Security\Authentication\Token;
 
+use Deprecated;
 use InvalidArgumentException;
 use LogicException;
 use RuntimeException;
@@ -88,7 +89,7 @@ class TwoFactorToken implements TwoFactorTokenInterface
         return $this->credentials;
     }
 
-    #[\Deprecated]
+    #[Deprecated]
     public function eraseCredentials(): void
     {
         $this->credentials = null;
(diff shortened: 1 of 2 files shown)


scheb:
Sure, why not. Released under v7.10.1.

---

Though may I ask a question on that occasion, and please don't take this offensive, I just want to understand. Why are deprecation messages such a big deal, that people are issuing bugs and requesting fixes for it? I mean, your application is not breaking from it. As long as the change will be addressed in a future release, when compatibility with the breaking change is _necessary_, we should be all good. Then I don't really understand, why people insist that deprecations are handled right away, when the software is just working fine. The only explanation that I have is, it bothers people having deprecation warnings in their logs.

If you could enlighten me, why you felt this needs to be changed, would be much appreciated :)

jkgroupe:
> Sure, why not. Released under v7.10.1.
> 
> Though may I ask a question on that occasion, and please don't take this offensive, I just want to understand. Why are deprecation messages such a big deal, that people are issuing bugs and requesting fixes for it? I mean, your application is not breaking from it. As long as the change will be addressed in a future release, when compatibility with the breaking change is _necessary_, we should be all good. Then I don't really understand, why people insist that deprecations are handled right away, when the software is just working fine. The only explanation that I have is, it bothers people having deprecation warnings in their logs.
> 
> If you could enlighten me, why you felt this needs to be changed, would be much appreciated :)

I didn't want it fixed immediately, I just wanted to offer my help to fix the issues. 
I opened a ticket but I didn't know if it would be fixed before the next Symfony release so I created a PR. 

But don't worry, there was no rush ;) 

I'll make sure to be more careful next time

PR #295, opened 2025-10-14
Add validator for a user's TOTP code

codedmonkey:
**Description**

Add a Symfony validator constraint to check the user's TOTP code. Based on the [`UserPassword`](https://symfony.com/doc/current/reference/constraints/UserPassword.html) constraint from Symfony.

This allows developers to integrate a TOTP check in Symfony forms without having to manually validate the code (including the setup form when first activating TOTP for the account), simply by adding the `UserTotpCode` constraint to the Symfony form, the entity (as an attribute) or by calling the Symfony validator itself.

I only a created it for TOTP because that's what I'm using in a project, but I'm willing to help build it for other 2FA authentication methods if this one gets accepted.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
         "psr/container": ">=1.1",
         "squizlabs/php_codesniffer": "^4.0",
         "symfony/mailer": "^6.4 || ^7.0",
+        "symfony/validator": "^6.4 || ^7.0",
         "symfony/yaml": "^6.4 || ^7.0",
         "vimeo/psalm": "^6.0"
     },
--- a/src/google-authenticator/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCode.php
+++ b/src/google-authenticator/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCode.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Validator\Constraints;
+
+use Attribute;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Validator constraint for the current user's Google Authenticator TOTP code.
+ *
+ * @final
+ */
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+class UserGoogleTotpCode extends Constraint
+{
+    public const INVALID_TOTP_CODE_ERROR = '6de3acd0-12f5-40eb-a776-2525c4566649';
+
+    protected const ERROR_NAMES = [self::INVALID_TOTP_CODE_ERROR => 'INVALID_TOTP_CODE_ERROR'];
+
+    public string $message = 'code_invalid';
+    public string $translationDomain = 'SchebTwoFactorBundle';
+    public string $service = 'scheb_two_factor.security.totp.validator.user_google_totp_code';
+
+    public function __construct(array|null $options = null, string|null $message = null, string|null $translationDomain = null, string|null $service = null, array|null $groups = null, mixed $payload = null)
+    {
+        parent::__construct($options, $groups, $payload);
+
+        $this->message = $message ?? $this->message;
+        $this->translationDomain = $translationDomain ?? $this->translationDomain;
+        $this->service = $service ?? $this->service;
+    }
+
+    public function validatedBy(): string
+    {
+        return $this->service;
+    }
+}
--- a/src/google-authenticator/composer.json
+++ b/src/google-authenticator/composer.json
@@ -16,6 +16,9 @@
         "scheb/2fa-bundle": "self.version",
         "spomky-labs/otphp": "^11.0"
     },
+    "suggest": {
+        "symfony/validator": "Needed if you want to use the Google Authenticator TOTP validator constraint"
+    },
     "autoload": {
         "psr-4": {
             "Scheb\\TwoFactorBundle\\": ""
--- a/src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCode.php
+++ b/src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCode.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Validator\Constraints;
+
+use Attribute;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Validator constraint for the current user's TOTP code.
+ *
+ * @final
+ */
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+class UserTotpCode extends Constraint
+{
+    public const INVALID_TOTP_CODE_ERROR = 'f79333fd-6eef-4394-ab6b-54827e2865b6';
+
+    protected const ERROR_NAMES = [self::INVALID_TOTP_CODE_ERROR => 'INVALID_TOTP_CODE_ERROR'];
+
+    public string $message = 'code_invalid';
+    public string $translationDomain = 'SchebTwoFactorBundle';
+    public string $service = 'scheb_two_factor.security.totp.validator.user_totp_code';
+
+    public function __construct(array|null $options = null, string|null $message = null, string|null $translationDomain = null, string|null $service = null, array|null $groups = null, mixed $payload = null)
+    {
+        parent::__construct($options, $groups, $payload);
+
+        $this->message = $message ?? $this->message;
+        $this->translationDomain = $translationDomain ?? $this->translationDomain;
+        $this->service = $service ?? $this->service;
+    }
+
+    public function validatedBy(): string
+    {
+        return $this->service;
+    }
+}
--- a/src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCodeValidator.php
+++ b/src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCodeValidator.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Validator\Constraints;
+
+use Scheb\TwoFactorBundle\Model\Totp\TwoFactorInterface;
+use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Totp\TotpAuthenticatorInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationCredentialsNotFoundException;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+use function get_debug_type;
+use function is_string;
+use function sprintf;
+
+/**
+ * Validator for the `UserTotpCode` constraint.
+ *
+ * @final
+ *
+ * @psalm-suppress PropertyNotSetInConstructor
+ */
+class UserTotpCodeValidator extends ConstraintValidator
+{
+    public function __construct(
+        private readonly TokenStorageInterface $tokenStorage,
+        private readonly TotpAuthenticatorInterface $totpAuthenticator,
+    ) {
+    }
+
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof UserTotpCode) {
+            throw new UnexpectedTypeException($constraint, UserTotpCode::class);
+        }
+
+        if (null === $value || '' === $value) {
+            $this->context->buildViolation($constraint->message)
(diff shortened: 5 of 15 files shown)


composer.json
```diff
@@ -36,6 +36,7 @@
         "psr/container": ">=1.1",
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/mailer": "^6.4 || ^7.0",
+        "symfony/validator": "^6.4 || ^7.0",
```
scheb, line 39: This line also needs to be added in the `src/totp/composer.json` as a dependency of the sub-package. If we add a validator for the Google Authenticator method, same for `src/google/composer.json`.

src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCode.php
```diff
@@ -0,0 +1,34 @@
...
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Validator\Constraints;
+
+use Attribute;
+use Symfony\Component\Validator\Constraint;
+
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+final class UserTotpCode extends Constraint
```
scheb: Please use annotation `@final` instead of the final keyword, as I've consistently used that approach all over the bundle. I might change it at some point, but that's something for the future.

(Should be changed for all classes in `src`)

src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCodeValidator.php
```diff
@@ -0,0 +1,70 @@
...
+                ->setCode(UserTotpCode::INVALID_TOTP_CODE_ERROR)
+                ->setTranslationDomain($constraint->translationDomain)
+                ->addViolation();
+
+            return;
+        }
+
+        if (!is_string($value)) {
+            throw new UnexpectedTypeException($value, 'string');
+        }
```
scheb, line 51: Not sure, if we have to be that strict. I'd probably just stringify the value and let the code validation fail. What was your thoughts on this?

tests/Security/TwoFactor/Validator/Constraints/UserTotpCodeTest.php
```diff
@@ -0,0 +1,59 @@
...
+use Generator;
+use PHPUnit\Framework\Attributes\DataProvider;
+use Scheb\TwoFactorBundle\Security\TwoFactor\Validator\Constraints\UserTotpCode;
+use Scheb\TwoFactorBundle\Tests\TestCase;
+use Symfony\Component\Validator\Mapping\ClassMetadata;
+use Symfony\Component\Validator\Mapping\Loader\AttributeLoader;
+
+class UserTotpCodeTest extends TestCase
+{
+    public function testValidatedByStandardValidator(): void
```
scheb: All test method names should follow `[method]_[scenario]_[expectedResult]` naming pattern (see [CONTRIBUTING.md](https://github.com/scheb/2fa/blob/7.x/CONTRIBUTING.md#creating-a-pull-request))

tests/Security/TwoFactor/Validator/Constraints/UserTotpCodeTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $constraint = new UserTotpCode();
+
+        self::assertSame('scheb_two_factor.security.totp.validator.user_totp_code', $constraint->validatedBy());
```
scheb: For PHPUnit helpers, please use `$this->*` instead of `self::*`. I know, it's the same, but I'd like to be consistent in the codebase.

(This and the other test class)

scheb (changes requested):
Looks really good, very thorough testing, thanks for going the effort!

If we add this, we should also add it for the Google Authenticator method to provide consistent solutions to the users.

And we should add some documentation on the validator(s).

composer.json
codedmonkey: I can but it’s an optional feature, so it’s only a dev dependency and should only be a suggested install imo. And since none of the other packages have any third-party dependencies right now I thought you wouldn’t want it for this one either.

The mail package is even missing the `symfony/mailer` dependency.

scheb: For the email-package it was done like that, because it doesn't want to force people to use the `symfony/mailer` package for email sending. So in the installation instructions it mentions to install that package, when you want to use the default mail-sending implementation: https://symfony.com/bundles/SchebTwoFactorBundle/7.x/providers/email.html#installation

I get the point making it an optional feature. I think the best way would be to add a "suggest" in the `composer.json` files, to point out the dependency to `symfony/validator` when the validation constraints are used.

Since you mentioned the email package, I just added that for the `symfony/mailer` dependency -> https://github.com/scheb/2fa/commit/6979b2aecff94cf6d43247f89c44a7afce63dcfa

scheb:
Hey @codedmonkey, I've seen you added some more changes. Are you planning to move forward and finish the changes necessary for the merge? I was just about to pick up where you left, because I would like to include this in the upcoming v8 release.

src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCodeValidator.php
codedmonkey: I just followed [UserPasswordValidator](https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/Security/Core/Validator/Constraints/UserPasswordValidator.php) and I think it's a good idea to enforce it because for numeric codes we don't want the confuse the developer by accidentally stripping any leading zeroes.

tests/Security/TwoFactor/Validator/Constraints/UserTotpCodeTest.php
codedmonkey: I wasn't able to change the one on https://github.com/scheb/2fa/pull/295/files#diff-6eb003c666580586a2490cf9a3c2abc45579730e1c275719fafaeb3f1923cad2R42 because it's in a static function

tests/Security/TwoFactor/Validator/Constraints/UserTotpCodeTest.php
codedmonkey: Changed, but `attributes_loadMetdata_returnValid` doesn't make a lot of sense on https://github.com/scheb/2fa/pull/295/files#diff-6eb003c666580586a2490cf9a3c2abc45579730e1c275719fafaeb3f1923cad2R48

codedmonkey:
Yeah sorry about the wait. Hoping I can finish it this week. I implemented all your feedback for the TOTP validator so let me know if there's still something missing.

src/totp/Security/TwoFactor/Validator/Constraints/UserTotpCodeValidator.php
scheb: Okay, then let's keep it like this 👍

src/google-authenticator/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCode.php
```diff
@@ -0,0 +1,39 @@
...
+
+declare(strict_types=1);
+
+namespace Scheb\TwoFactorBundle\Security\TwoFactor\Validator\Constraints;
+
+use Attribute;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Validator constraint for the current user's Google TOTP code.
```
scheb: To be precise:
```suggestion
 * Validator constraint for the current user's Google Authenticator TOTP code.
```

src/google-authenticator/composer.json
```diff
@@ -16,6 +16,9 @@
         "scheb/2fa-bundle": "self.version",
         "spomky-labs/otphp": "^11.0"
     },
+    "suggest": {
+        "symfony/validator": "Needed if you want to use the TOTP validator constraint"
```
scheb: To be precise:
```suggestion
        "symfony/validator": "Needed if you want to use the Google Authenticator TOTP validator constraint"
```

tests/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCodeValidatorTest.php
```diff
@@ -0,0 +1,157 @@
...
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+use Symfony\Component\Validator\Test\ConstraintValidatorTestCase;
+
+/**
+ * @implements ConstraintValidatorTestCase<UserGoogleTotpCodeValidator>
+ */
+class UserGoogleTotpCodeValidatorTest extends ConstraintValidatorTestCase
```
scheb, line 24: Tests are looking good. Just some nitpicking about the declared method/attribute visibility. Most of the attributes and methods should be private rather than protected. Except it's a method/attribute overwritten from ConstraintValidatorTestCase.

Same with the UserTotpCodeValidatorTest (copy & paste, I guess).

tests/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCodeTest.php
```diff
@@ -0,0 +1,63 @@
...
+    public static function provideServiceValidatedConstraints(): iterable
+    {
+        yield 'Doctrine style' => [new UserGoogleTotpCode(['service' => 'my_service'])];
+
+        yield 'named arguments' => [new UserGoogleTotpCode(service: 'my_service')];
+
+        $metadata = new ClassMetadata(UserTotpCodeDummy::class);
+        self::assertTrue((new AttributeLoader())->loadClassMetadata($metadata));
+
+        yield 'attribute' => [$metadata->getPropertyMetadata('b')[0]->getConstraints()[0]];
```
scheb: This doesn't feel right to me, having an assert in a _data provider_, which should return _static_ values to represent different test scenarios. As soon as there's logic and asserts involved, the data provider is no longer a data provider, but becomes a test case.

I'd rather move the check for the "validatedBy" value to the `attributes_loadMetadata_returnValid` test. To my understanding, this is a "check values loaded from attributes are properly processed" test case, so would be better to assert the attribute-provided validatedBy value there.

(same in the UserTotpCodeTest)

tests/Security/TwoFactor/Validator/Constraints/UserTotpCodeTest.php
scheb: Resolving this thread. I'll add a comment on that static case in a second.

tests/Security/TwoFactor/Validator/Constraints/UserTotpCodeTest.php
scheb: Yea, that test case is special, since you don't have a specific "method under test" in that scenario, so naming becomes a bit odd.

In that case I might just skip the "method" part and go with something like: `configureConstraintFromAttribute_configurationIsCorrect`

scheb:
You'd want to rebase onto current 7.x branch, to fix the code style issues that are not part of your code changes :)

tests/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCodeTest.php
codedmonkey: I rewrote the entire test file to only the method that checks the attributes since it already does what the other tests do. What do you think?

codedmonkey:
I asked AI to generate some docs but it obviously not fully usable but I liked to have a direction to take the documentation, see https://github.com/codedmonkey/scheb-2fa/commits/totp-validator-ai-slop/

Let me know if you have some ideas on how the docs should look.

scheb:
AI going completely overboard as usual 😁.

In my opinion the "Available Validation Options" part would be sufficient as an explanation. Show the usage of the validation constraints with custom error messages, good enough. We don't have to explain what validators are, features like groups, etc. That's what the official documentation is for.

Would keep the intro sentence, as it explains the use-case for these specific validator constraints:

> When building forms that require users to enter two-factor authentication codes, you can use the built-in validation constraints provided by SchebTwoFactorBundle to automatically verify the codes.

Where documentation for the constraints should be located:

- Either keep it as you have it right now in its own file. Add a link in the `index.rst` and have it linked in `totp.rst` and `google.rst` (like the "Events" section there)
- Or split it up, to have a new section for each constraint respectively in `totp.rst` and `google.rst`

tests/Security/TwoFactor/Validator/Constraints/UserGoogleTotpCodeTest.php
scheb: Works for me 👍

scheb:
So far looks good to me. Docs need to be added.

PHP-CS-Fixer and PHP-Codesniffer can't agree on the empty newline after `@final` in those two files 🤷. I've excluded the offending rule on the 7.x branch (https://github.com/scheb/2fa/commit/06936f161cd2173e16988dcf409c57118855264f), so nothing to worry about.

scheb:
I'll wrap this up myself and add the docs, because I want to get the v8 out. Thanks for the hard work! Much appreciated! 🙇

codedmonkey:
Thanks for merging and your work as well 🎉

PR #298, opened 2025-11-20
ci: bump doctrine/coding-standard and squizlabs/php_codesniffer

Chris53897:


scheb:
Doesn't seem to break the code style checks. Sure, we can upgrade those

Issue #270, opened 2025-03-27
Bundle version 8 checklist


scheb:
This issue is to keep track of my todo list for the upcoming bundle version 8 (Symfony 8 support) 😀

The plan is, as usual:
- A new version of the bundle is released shortly after Symfony 8 is officially released. This typically happens in the same time frame when a new PHP version is released as well. So please be patient, until work is finished.
- A `8.x` branch will eventually appear, with an ⚠unstable development state. It is not recommended to be used in production scenarios, as there will a lot of changes and rebasing on this branch. Use it at your own risk.

---

Features

- [x] Change security factory priority to < -50 (-100?) -> #265

Codebase upgrades

- [x] Remove deprecated properties/methods
- [x] Upgrade codebase to PHP 8.x (min PHP version TBD) style where possible
- [x] Upgrade code examples in the docs with PHP 8.x (min PHP version TBD) code style
- [x] Upgrade notes
- [x] Clean up code that is no longer needed and remove compatibility layers
- [x] Update CONTRIBUTING.md on language features to be used
- [x] Update .symfony.bundle.yaml
- [x] Update version number and compatible Symfony versions in 8.x docs
- [x] Drop old versions of dependencies
- [x] Update mentions of "7.x" to "8.x" on the 8.x branch (i.e. docs and issue templates and gitsplit workflow)
- [x] Update version guidance on 8.x branch
- [x] Update "replace" section in composer.json
- [x] Clean up the `app`
- [x] Upgrade code style
- [x] Check for skipped tests, whether these are still needed
- [x] Working integration tests
- [x] Update gitsplit config
- [x] Update Pslam PHP version in `psalm.xml`
- [x] Update PHP-CS-Sniffer PHP version in `phpcs.xml.dist`
- [x] `MockObject|` -> `MockObject&`

Pre-Release Cleanup (once Symfony 8 stable is released)

- [x] Remove `minimum-stability` and `prefer-stable` from composer.json files
- [x] Remove temporary changes from composer.json files, i.e. `@dev`
- [x] Rewrite commits to clean up fiddling with CI and dependencies

Release

- [x] Change default branch on Github to 8.x
- [x] Release 🥳 
- [x] Abandon version 6.x, update version guidance
- [x] Update version guidance on 7.x branch

Post-Release

- [x] Remove `.symfony.bundle.yaml` from 7.x
- [x] Change default branch of sub-packages to 8.x

alexander-schranz:
[...]

PR #300, opened 2025-11-20
feat: allow symfony 8 (WIP)

Chris53897:
https://github.com/scheb/2fa/issues/270

WIP

need to be reverted before merging.
`"minimum-stability": "rc",
    "prefer-stable": false`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,6 +72,10 @@ jobs:
                     -   php-version: '8.5'
                         symfony-version: '7.*'
 
+                    # Test against Symfony 8 on highest PHP version
+                    -   php-version: '8.5'
+                        symfony-version: '8.*'
+
         steps:
             -   name: 'Checkout code'
                 uses: actions/checkout@v5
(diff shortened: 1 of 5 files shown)


scheb:
Thanks for giving this a shot. I got a bit confused when I saw so many old version builds failing and got even more confused when I saw that the 6.4 build had Symfony 8 dependencies installed. Turned out there was a caching issue in the composer-install step, that caused dependencies from other build jobs leaking in.

That got fixed on the 7.x branch now, so please rebase to give it another shot. Though you should be warned, from my experience, juggling half-cooked Symfony 8 dependencies to find a working combination will be no easy endeavor 🙈

Chris53897:
Thanks for fixing the ci.
I will upgrade our company project to PHP 8.5 and Symfony 8.0. So there are many libraries to upgrade at the moment.
I try to find a solution for this package, but not sure if i can fix it.

scheb:
The [8.x development branch](https://github.com/scheb/2fa/tree/8.x) is now building successfully against Symfony 7.4 and 8.0 development versions. Feel free to use it, but be aware that I'll mess around with it a lot (rebasing etc.) until it's ready for release.

Chris53897:
With this version constraints it will get easier to support Symfony 7.4 and 8.0.
I like it

PR #302, opened 2025-11-29
Allow Symfony 8

slaubi:
Symfony just released version 8. can we bump the version?


scheb:
Please see https://github.com/scheb/2fa/issues/270, the 8.x is branch is what you're looking for.

PR #307, opened 2026-01-12
Allow TwoFactor providers to be stateless

### LINKED ISSUE #306: Add support for 2FA in applications without state
Currently, it is not possible to use stateless 2fa methods, like TOTP or Google Authenticatior without a session.
In the linked issue / PR, this was implemented and merged into 8.x.  
However, these changes did not make it into the final release.

Prior discussion:
https://github.com/scheb/2fa/pull/273
https://github.com/scheb/2fa/issues/265

Can we re-introduce these changes, so we can use 2fa in a stateless app?

PR: https://github.com/scheb/2fa/pull/307
PR: https://github.com/scheb/2fa/pull/308

johanib:
Prior to this change, there was no way to determine if a two-factor provider needed preparation before authentication.

This change introduces the needsPreparation method in the TwoFactorProviderInterface and its implementations, allowing the system to skip the preparation process for providers that do not require it. The preparation process requires state.

For example:
Prior to this change, if no state was available, the  Totp and Google authenticators would fail, even if they are stateless.

Fixes: https://github.com/scheb/2fa/issues/306

This PR introduces backward incompatible changes, as custom implementations need to implement the additional method.

### DIFF
--- a/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php
+++ b/src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInitiator.php
@@ -8,6 +8,8 @@ use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenFactoryInt
 use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\AuthenticationContextInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Exception\UnknownTwoFactorProviderException;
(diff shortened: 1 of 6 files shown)


scheb:
Bummer, it must have somehow gotten lost during all those rebase/rewrite operations on the 8.x branch :(

Happy to introduce it back in, but then the introduction of the `needsPreparation` method would be a breaking change for the `TwoFactorProviderInterface`. So the way forward would be to "soft-introduce" it with an annotation and have checks, whether the method is declared - otherwise assume it `false` to return to the old behavior.

johanib:
I added a backwards compatible version in https://github.com/scheb/2fa/pull/308

PR #308, opened 2026-01-13
Allow TwoFactor providers to be stateless

### LINKED ISSUE #306: Add support for 2FA in applications without state

johanib:
Prior to this change, there was no way to determine if a two-factor provider needed preparation before authentication.

This change introduces the needsPreparation method in the TwoFactorProviderInterface and its implementations, allowing the system to skip the preparation process for providers that do not require it. The preparation process requires state.

For example:
Prior to this change, if no state was available, the  Totp and Google authenticators would fail, even if they are stateless.

Now, non-bc breaking.

Fixes https://github.com/scheb/2fa/issues/306


src/bundle/Security/TwoFactor/Provider/TwoFactorProviderInterface.php
```diff
@@ -13,6 +13,15 @@ interface TwoFactorProviderInterface
      */
     public function beginAuthentication(AuthenticationContextInterface $context): bool;
 
+    /**
+     * Determine whether this Provider needs to be prepared (if the prepareAuthentication method needs to be called).
+     *
+     * In version 9, this method will be introduced, and all providers will need to implement it.
+     * Currently, it will be called, but is not required.
+     *
+     * public function needsPreparation(): bool;
```
scheb, line 25: Please add `@method bool needsPreparation()` annotation on the interface.

tests/Security/TwoFactor/Provider/TwoFactorProviderInitiatorTest.php
```diff
@@ -11,22 +11,21 @@
...
+    private MockTwoFactorProvider $provider2;
     private MockObject&TwoFactorProviderDeciderInterface $providerDecider;
     private TwoFactorProviderInitiator $initiator;
 
     protected function setUp(): void
     {
-        $this->provider1 = $this->createMock(TwoFactorProviderInterface::class);
-        $this->provider2 = $this->createMock(TwoFactorProviderInterface::class);
+        $this->provider1 = new MockTwoFactorProvider(needsPreparation: true);
+        $this->provider2 = new MockTwoFactorProvider(needsPreparation: false);
```
scheb: If this is to introduce the `needsPreparation` method, the better thing you could do is a `TwoFactorProviderWithNeedsPreparationInterface extends TwoFactorProviderInterface` to "materialize" the method. Then you can use normal PHPUnit mocking. No need for custom assertion/return value stubbing in that case. The less custom code, the better.

Would also introduce a 3rd provider that is a mock of `TwoFactorProviderInterface` without the `needsPreparation` method, which should behave like the `needsPreparation` returning `true` case.

scheb (changes requested):
Thanks for going through the effort adjusting the implementation. I believe we can improve the test case a bit. Other than that, I'm happy to merge that in.

tests/Security/TwoFactor/Provider/TwoFactorProviderInitiatorTest.php
johanib: :heavy_check_mark: 
The mocks were getting confused, but the extra interface solves that :+1: 

It took some time to understand the provider1 / provider2 difference, so I took the liberty to rename them..

I hope this now matches your expectations. Breaking the check in `beginTwoFactorAuthentication` always results in a red pipeline.

scheb: Much better with the descriptive provider names, thanks! 👍

scheb:
Released as v8.3.0

PR #310, opened 2026-02-09
Fix version number (v8) in installation instructions

michael-bouvy:
Thanks for the great job. This tiny MR only fixes the version number in installation instructions for v8.x.


scheb:
Thanks!

PR #311, opened 2026-02-23
Align TOTP configuration on Attributes/Annotations examples

thePanz:
**Description**
Align the example changes for both Attributes and Annotations example on the User entity:
- check for secret existence before building the Configuration class
- use a configuration compatible with Google-Authenticator (digits:6, period: 30) in both examples

### DIFF
--- a/doc/providers/totp.rst
+++ b/doc/providers/totp.rst
@@ -94,7 +94,7 @@ and the period of the temporary codes.
            public function getTotpAuthenticationConfiguration(): ?TotpConfigurationInterface
            {
                // You could persist the other configuration options in the user entity to make it individual per user.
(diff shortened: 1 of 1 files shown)


scheb:
No thanks. The whole point of these values is to show that you can have whole different configurations than the 6/30 that Google Authenticator has popularized.

If you want Google Authenticator compatible settings, no need to use the `scheb/2fa-totp` package, just use the `scheb/2fa-google-authenticator` one.

thePanz:
Thanks @scheb for the reply. I agree that calling "Google" the defaults for the [RFC-6238](https://www.rfc-editor.org/rfc/rfc6238) has been too much popularized :)

PR #312, opened 2026-02-24
fix: stateless 2fa provider preparation listener

dt-thomas-durand:
**Description**
In certain scenarios, TwoFactorProviderPreparationListener might not be re-instanciated between requests. 
That's the case when multiple requests are handled by a single worker, like when serving the application with RoadRunner or with FrankenPHP. 
In that case, the twoFactorToken value might leak from a request to another, leading to unwanted preparations.
By implementing ResetInterface, the token is now properly reseted between requests


scheb:
Thanks, released under v8.4.0

PR #315, opened 2026-03-23
feat: added condition skipping 2fa event

dt-thomas-durand:
**Description**
Added a new SKIPPED case to `TwoFactorAuthenticationEvents` ; triggered with a TwoFactorAuthenticationEvent object when Two Factor is skipped because a condition returned false


scheb:
Looks good! Would you please add the event in the docs? Then we can merge.

dt-thomas-durand:
> Looks good! Would you please add the event in the docs? Then we can merge.

Documentation updated!

scheb:
Released as v8.5.0

PR #316, opened 2026-05-06
Close #213: honor post-success listeners that may block authentication

### LINKED ISSUE #213: Infinite redirection loop problem
**Bundle version**: 7.0.0
**Symfony version**: 7.0.1
**PHP version**: 8.2.10

**Description**

Throwing an exception in `user_checker` `checkPostAuth()` ends in infinite redirection loop.

**To Reproduce**

0. Configure bundle using email tokens.
1. Add user checker in `security.yaml`
```yaml
security:
  firewalls:
    website:
      lazy: true
      provider: app_user_provider
      user_checker: App\Security\UserChecker
      ...
```
2. `App\Security\UserChecker`
```php
<?php

namespace App\Security;

use Symfony\Component\Security\Core\Exception\CredentialsExpiredException;
use Symfony\Component\Security\Core\User\UserCheckerInterface;
use Symfony\Component\Security\Core\User\UserInterface;

class UserChecker implements UserCheckerInterface
{
    public function checkPreAuth(UserInterface $user): void
    {
    }

    public function checkPostAuth(UserInterface $user): void
    {
        throw new CredentialsExpiredException();
    }
}
```

3. Now login as usual, enter the code from the mail, submit form and you will get the infinite redirection loop.

image

giosh94mhz:
### Description

Symfony’s standard authentication includes several "post-checks" that run immediately after a successful login, most notably `TwoFactorAuthenticator::onAuthenticationSuccess`.

Starting with Symfony 7.x and 2fa-bundle 7.x (and possibly others), these listeners can generate a loop-redirect, as described in #213.

### The Issue

The root cause is that `TwoFactorToken` marks a provider as completed during badge verification. If an error occurs after this stage, the entire login process is left in an inconsistent state for several reasons:

1. There is no "current provider," even if the system is still processing a provider (e.g., TOTP).
2. At this point, the stored providers (i.e., `getTwoFactorProviders`) are empty, so the active token in `TokenStorage` is inconsistent
3. If authentication fails, the current authentication token is  already switched and the previous inconsistent token remains in storage.
4. Subsequent requests then generate a new "prepared providers" list, but no providers are available, triggering the failure loop.

### The Solution

During badge verification, providers **must not** be marked as complete in the `TwoFactorToken`, as this happens too early in the process. This is critical to keep `TwoFactorTokenInterface::getCurrentTwoFactorProvider` stable throughout the request and avoid the loop issue.

Instead, we need to identify another point to mark the `TwoFactorToken` provider as complete—ideally, at the very end of the authentication process.

By examining the `Symfony\Component\Security\Http\Authentication\AuthenticatorManager::executeAuthenticator` class, I see that the `AuthenticationEvents::AUTHENTICATION_SUCCESS` event (where `UserCheckers` and other checks run) is considered part of the standard process, as it is within the try-catch block. Immediately after this event, `$authenticator->onAuthenticationSuccess` is called, making it the right place to mark the flow as completed.

**Therefore, the solution is to mark the current provider for any `TwoFactorToken` that reaches `TwoFactorAuthenticator::onAuthenticationSuccess` as completed—not before.**

### Tests

I ran many tests in my application, which uses both stateful and stateless authenticators, as well as multiple post-authentication locks, and everything seems to work. However, we only have one OTP method, so the code for handling multiple OTP methods and `$this->dispatchTwoFactorAuthenticationEvent(TwoFactorAuthenticationEvents::REQUIRE, $request, $token)` still needs some testing.

### Minor Breaking Changes

I foresee two minor breaking changes:

1. Updating users may experience changes in `TwoFactorTokenInterface::getCurrentTwoFactorProvider`, e.g. from `null` to `totp` (though this is likely a fix rather than an issue, IMO).
2. The `TwoFactorTokenInterface::allTwoFactorProvidersAuthenticated` method will likely never return `true`, since the previous token forgot after `onSuccessfulAuthentication`.

The main struggle with `TwoFactorTokenInterface` is that the `setTwoFactorProviderComplete` method actually removes a provider from the token, which can cause glitches. My patch addresses this by requiring the authenticator to estimate the "on complete" status by comparing `getTwoFactorProviders` and the active providers. 

Regarding point 2, I may update the `TwoFactorToken` inside the authenticator success so that `TwoFactorTokenInterface::allTwoFactorProvidersAuthenticated` is updated for potential external classes still referencing the token and triggering in post-success events, like `TwoFactorAuthenticationEvents::COMPLETE`.

In a future release, `TwoFactorTokenInterface` may need a cleanup or an improvements to avoid this edge cases.

### DIFF
--- a/src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
+++ b/src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
@@ -29,17 +29,22 @@ use Symfony\Component\Security\Http\Authenticator\Passport\Badge\RememberMeBadge
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\Service\ResetInterface;
 use function assert;
 use function class_exists;
+use function count;
+use function spl_object_hash;
 
 /**
  * @final
  */
-class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthenticatorInterface
+class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthenticatorInterface, ResetInterface
 {
     public const FLAG_2FA_COMPLETE = '2fa_complete';
 
     private readonly LoggerInterface $logger;
+    /** @var TwoFactorTokenInterface[] */
+    private array $onCompleteTokens = [];
 
     public function __construct(
         private readonly TwoFactorFirewallConfig $twoFactorFirewallConfig,
@@ -51,6 +56,11 @@ class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthe
         LoggerInterface|null $logger = null,
     ) {
         $this->logger = $logger ?? new NullLogger();
+    }
+
+    public function reset(): void
+    {
+        $this->onCompleteTokens = [];
     }
 
     public function supports(Request $request): bool|null
@@ -113,6 +123,7 @@ class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthe
         if ($this->isAuthenticationComplete($twoFactorToken)) {
             $authenticatedToken = $twoFactorToken->getAuthenticatedToken(); // Authentication complete, unwrap the token
             $authenticatedToken->setAttribute(self::FLAG_2FA_COMPLETE, true);
+            $this->onCompleteTokens[spl_object_hash($authenticatedToken)] = $twoFactorToken;
 
             return $authenticatedToken;
         }
@@ -122,7 +133,10 @@ class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthe
 
     private function isAuthenticationComplete(TwoFactorTokenInterface $token): bool
     {
-        return !$this->twoFactorFirewallConfig->isMultiFactor() || $token->allTwoFactorProvidersAuthenticated();
+        return !$this->twoFactorFirewallConfig->isMultiFactor()
+            || $token->allTwoFactorProvidersAuthenticated()
+            // The current provider is the last provider. It will be completed in onAuthenticationSuccess.
+            || 1 === count($token->getTwoFactorProviders());
     }
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): Response|null
@@ -132,9 +146,23 @@ class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthe
 
         // When it's still a TwoFactorTokenInterface, keep showing the auth form
         if ($token instanceof TwoFactorTokenInterface) {
+            $currentProvider = $token->getCurrentTwoFactorProvider();
+            if (null !== $currentProvider) {
+                $token->setTwoFactorProviderComplete($currentProvider);
+            }
+
             $this->dispatchTwoFactorAuthenticationEvent(TwoFactorAuthenticationEvents::REQUIRE, $request, $token);
 
             return $this->authenticationRequiredHandler->onAuthenticationRequired($request, $token);
+        }
+
+        $twoFactorToken = $this->onCompleteTokens[spl_object_hash($token)] ?? null;
+        if (null !== $twoFactorToken) {
+            unset($this->onCompleteTokens[spl_object_hash($token)]);
+            $currentProvider = $twoFactorToken->getCurrentTwoFactorProvider();
+            if (null !== $currentProvider) {
+                $twoFactorToken->setTwoFactorProviderComplete($currentProvider);
+            }
         }
 
         $this->dispatchTwoFactorAuthenticationEvent(TwoFactorAuthenticationEvents::COMPLETE, $request, $token);
--- a/src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
+++ b/src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
@@ -123,6 +123,9 @@ class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthe
         if ($this->isAuthenticationComplete($twoFactorToken)) {
             $authenticatedToken = $twoFactorToken->getAuthenticatedToken(); // Authentication complete, unwrap the token
             $authenticatedToken->setAttribute(self::FLAG_2FA_COMPLETE, true);
+
+            // Backwards compatibility for bundle version <=8
+            // Remember the token to set the provider complete in onAuthenticationSuccess
             $this->onCompleteTokens[spl_object_hash($authenticatedToken)] = $twoFactorToken;
 
             return $authenticatedToken;
@@ -156,13 +159,17 @@ class TwoFactorAuthenticator implements AuthenticatorInterface, InteractiveAuthe
             return $this->authenticationRequiredHandler->onAuthenticationRequired($request, $token);
         }
 
+        // Backwards compatibility for bundle version <=8
+        // In case the two-factor process was completed, $token is no longer a TwoFactorToken
+        // Set the provider completed on the remembered TwoFactorToken
         $twoFactorToken = $this->onCompleteTokens[spl_object_hash($token)] ?? null;
         if (null !== $twoFactorToken) {
-            unset($this->onCompleteTokens[spl_object_hash($token)]);
             $currentProvider = $twoFactorToken->getCurrentTwoFactorProvider();
             if (null !== $currentProvider) {
                 $twoFactorToken->setTwoFactorProviderComplete($currentProvider);
             }
+
+            unset($this->onCompleteTokens[spl_object_hash($token)]);
         }
 
         $this->dispatchTwoFactorAuthenticationEvent(TwoFactorAuthenticationEvents::COMPLETE, $request, $token);
--- a/tests/Security/Http/Authenticator/TwoFactorAuthenticatorTest.php
+++ b/tests/Security/Http/Authenticator/TwoFactorAuthenticatorTest.php
@@ -30,8 +30,10 @@ use Symfony\Component\Security\Http\Authenticator\Passport\Badge\CsrfTokenBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\RememberMeBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use function array_values;
 use function assert;
 use function method_exists;
+use function reset;
 
 class TwoFactorAuthenticatorTest extends TestCase
 {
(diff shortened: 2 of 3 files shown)


src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
```diff
@@ -110,6 +110,9 @@ public function createToken(Passport $passport, string $firewallName): TokenInte
         assert($credentialsBadge instanceof TwoFactorCodeCredentials);
         $twoFactorToken = $credentialsBadge->getTwoFactorToken();
 
+        // Provider complete can only be marked this onAuthenticationSuccess, since others auth listener may still trigger failures
+        // $twoFactorToken->setTwoFactorProviderComplete($twoFactorToken->getCurrentTwoFactorProvider());
```
scheb: I'm wondering where in the process the provider is marked as completed, if it's no longer happening in AbstractCheckCodeListener.

Should this line actually be un-commented?

giosh94mhz: You are right to wonder, and the answer is... actually never. (I tried to explain it in the issue, hopefully I passed the language barrier :smile: ) 

We can't uncomment that line, that's why I explicitly added it so it's not attempted in the future. If you do, you get the same looping bug, since "post-auth" lister are executed right after the token is switched.

The correct place is always `TwoFactorAuthenticator::onSuccessfulAuthentication`, but at that point we no longer have the original `TwoFactorToken` only the `SymfonyToken` (or whatever token). So we have three solution:

1. doing nothing, as the token will be forgotten right away (also, the `TokenStorage` no longer contains it)
2. keep a local map like "SymfonyToken => TwoFactorToken", so that we can retrieve `TwoFactorToken`
3. add a temporary attribute on SymfonyToken like "otp_completed_token = TwoFactorToken" , and reset it after use on success event

I've (not) done solution 1 for the sake of simplicity, but we may implement solution 3 (my favorite) or solution 2 (more fair with SymfonyToken) if you prefer.

-- edit fixed typo

scheb: I understand the problem and I also understand what you're trying to do.

I believe you implemented and tested this solution under the assumption that there is only _one_ provider that needs to pass. But if you have `multi_factor: true` activated on the firewall, _all_ available providers need to pass, before the MFA phase is completed. That's why it's necessary to flag providers as completed, so that MFA can proceed with the next provider, until all have passed. If you don't do that, you'll end up in an endless loop with `multi_factor: true` enabled. You can actually try that out by running the test application in the `app` folder. You'll see that you're stuck in an endless loop, asking to enter the code for the `email` provider over and over again.

Related, this statement is actually not true:

> The correct place is always TwoFactorAuthenticator::onSuccessfulAuthentication, but at that point we no longer have the original TwoFactorToken only the SymfonyToken (or whatever token).

You can indeed have a `TwoFactorToken` in this method. That happens when `multi_factor: true` is configured and the there are still providers left. See `TwoFactorAuthenticator.php:139`. So maybe the solution is to flag the provider as completed in this method?

---

Side note on the test application in `app`: This is also where integration tests are located. Your changes are breaking some integration tests. May be related to the problem that I've just described with `multi_factor: true`. In any case, I highly recommend to run the integration tests to validate to check for any regressions.

giosh94mhz: Sorry for the late reply, I prefer to answer with some code :smile: 

I've found how to run Symfony phpunit tests inside the project, didn't noticed the double configuration and so I missed the test. The failures were caused by a missing line in my patch I submitted, in my local code I had already set a provider as completed right before the "REQUIRE" event, thus fixing the whole multi-factor authentication. You were right, in the code I submitted an infinite loop will trigger.

Regarding the "on complete" status of the TwoFactorToken iteself, I've implemented in a separate commit the solution I described as "2", that is using a temporary attribute to remember the token and fixing it right before the event. Not very proud of this commit, should you prefer other solutions let me know.

giosh94mhz: ...and fixed all the tests again. Sorry for the noise.

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
```diff
@@ -122,7 +128,10 @@ public function createToken(Passport $passport, string $firewallName): TokenInte
 
     private function isAuthenticationComplete(TwoFactorTokenInterface $token): bool
     {
-        return !$this->twoFactorFirewallConfig->isMultiFactor() || $token->allTwoFactorProvidersAuthenticated();
+        return !$this->twoFactorFirewallConfig->isMultiFactor()
+            || $token->allTwoFactorProvidersAuthenticated()
+            // only one left, so it's "on complete"
+            || [(string) $token->getCurrentTwoFactorProvider()] === array_values($token->getTwoFactorProviders());
```
scheb: I'd rather go with approach `count($token->getTwoFactorProviders()) <= 1` to express "this is the last provider".

Also, I would comment  "The current provider is the last provider. It will be completed in onAuthenticationSuccess." to make the intention clear for future reference.

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
```diff
@@ -132,11 +141,21 @@ public function onAuthenticationSuccess(Request $request, TokenInterface $token,
...
             $this->dispatchTwoFactorAuthenticationEvent(TwoFactorAuthenticationEvents::REQUIRE, $request, $token);
 
             return $this->authenticationRequiredHandler->onAuthenticationRequired($request, $token);
         }
 
+        if ($token->hasAttribute(self::TOKEN_2FA_COMPLETE)) {
+            $twoFactorToken = $token->getAttribute(self::TOKEN_2FA_COMPLETE);
+            $twoFactorToken->setTwoFactorProviderComplete($twoFactorToken->getCurrentTwoFactorProvider());
+            $token->setAttribute(self::TOKEN_2FA_COMPLETE, null);
+        }
```
scheb, line 173: I agree, this part isn't pretty. Feels like a "hack".

I'd say we don't need this necessarily. When the 2fa token is completed, the `TwoFactorToken` is replaced by the authenticated token. So changing the "complete" state for the last provider is not really necessary, since we're ditching the `TwoFactorToken` anyways. I checked, if we're removing this part, 2fa is still working fine, also in the multi-factor case. The important part is to call `setTwoFactorProviderComplete` in line 145.

So I'd say we're going with your "option 1" for when the 2fa process is finally completed: 

> doing nothing, as the token will be forgotten right away (also, the `TokenStorage` no longer contains it)

This will introduce a this small behavioral change, that you have previously described:

> Updating users may experience changes in TwoFactorTokenInterface::getCurrentTwoFactorProvider, e.g. from `null` to `totp` (though this is likely a fix rather than an issue, IMO).

I'd be okay with that. Also agree, that this may rather be a "fix".

scheb (commented):
Overall, looks good, thanks for the effort!

Please remove commented out code fragments for the final, clean implementation.

src/bundle/Security/Http/Authenticator/TwoFactorAuthenticator.php
giosh94mhz: > Please remove commented out code fragments for the final, clean implementation.

I removed most of the comments and updated the one you pointed out

> `count($token->getTwoFactorProviders()) <= 1`

Had to check `1 === count($token->getTwoFactorProviders())` as some other test failed otherwise

> So I'd say we're going with your "option 1" for when the 2fa process is finally completed: [...]

Once I implemented the "token trick" I was no longer sure it was a good idea of doing nothing until the next "major". It's a minor BC, but I don't like to break other people code.

>> keep a local map like "SymfonyToken => TwoFactorToken", so that we can retrieve TwoFactorToken

I've done the above solution (option 2), that is fully encapsulated in `TwoFactorAuthenticator`.

IMO, we may flag method `allTwoFactorProvidersAuthenticated` as `@deprecated` so that in the next major you may remove it, and revert my 2nd commit with the trick.

What do you think?

scheb: I'm good with that. Will have another look, once I find some time.

scheb: Looks good, resolving this thread.

tests/Security/Http/Authenticator/TwoFactorAuthenticatorTest.php
```diff
@@ -361,10 +383,11 @@ public function createToken_noMultiFactorAuthentication_returnAuthenticatedToken
         $twoFactorToken = $this->createTwoFactorToken($authenticatedToken, false);
         $passport = $this->createPassportWithTwoFactorCredentials($twoFactorToken);
 
-        $this->expect2faCompleteFlagSet($authenticatedToken);
+        $this->stubAttributes($authenticatedToken);
 
         $returnValue = $this->authenticator->createToken($passport, self::FIREWALL_NAME);
         $this->assertSame($authenticatedToken, $returnValue);
+        $this->assertTrue($authenticatedToken->getAttribute(TwoFactorAuthenticator::FLAG_2FA_COMPLETE));
```
scheb: Please revert all those changes in `TwoFactorAuthenticatorTest`, since they're not necessary with the current implementation.

What I'd like to have instead, is test coverage for the `onCompleteTokens` being set and processed.

scheb (changes requested):

tests/Security/Http/Authenticator/TwoFactorAuthenticatorTest.php
giosh94mhz: Sure I can, but give it a double check first, since the test code is not working as intended IMO.

The previous code had the following:

```php
         $authenticatedToken
            ->expects($this->any())
             ->method('setAttribute')
            ->with(TwoFactorAuthenticator::FLAG_2FA_COMPLETE, true);
```

Since the expectation is "any" it doesn't really test the code. I mean, nothing is tested if setAttribute isn't called, but nothing is asserted too when FLAG_2FA_COMPLETE is set since this is always true. You are just testing that no other "setAttribute" are called, but I don't think that is what you meant.

At the very least, the previous test `expect2faCompleteFlagSet` should assert that "atLeastOne" setAttribute is called (IIRC I tested that at first and some other test failed... I may try again if you wish)

scheb: You can change that to `once` / `atLeastOnce`. I don't remember why it's `any`. But all the other changes should be reverted, as they're now unrelated to the current state of the PR and a leftover of the previous implementation.

giosh94mhz: Done! :+1:

scheb: Thanks, much cleaner now. Though that brings me back to my original comment:

> What I'd like to have instead, is test coverage for the `onCompleteTokens` being set and processed.

giosh94mhz: Ok, implementing that test was.... hard!

The problem I faced is that all the "main tests" assumes that you can tests every single authentication method as a standalone concept, while the "app tests" -- that were already testing that path -- are not accounted for the code coverage.

I found a clean way to test all the new code, hopefully is good enough, but IMO you should merge the two coverage as a future improvement.

scheb (commented):
**Regarding merging test coverage from integration tests:** The purpose of the integration tests is to validate the behavior of the system at a high level. They're not meant to validate implementation details. Therefore, only the test coverage from unit tests is used, as it gives a much clearer picture which parts of the code are actually tested in detail.

**Regarding the tests:** I'm happy with the general approach of calling `createToken` and `onAuthenticationSuccess` together in the same test to create a test scenario. That said, there are still a few aspects of the implementation that would need to be adjusted before this is ready to be merged. To help move things along, I'll take it from here and make the remaining changes myself, as I think that may be the quickest path forward. Thanks again for the work you’ve put into this.

What I specifically will change to polish this:

- `stubTwoFactorTokenProviders` feels every "vibe coded" the way it works. It contains too much logic for a stub. Stubs are supposed to be simple, ideally just return values.
- `stubTwoFactorTokenProviders` is used in `onAuthenticationSuccess_authenticationIncomplete_dispatchSuccessAndRequireEvent` and the whole test setup to assert `setTwoFactorProviderComplete` is there, but then nothing is asserted.
- Asserts for the `setTwoFactorProviderComplete` code path are mixed into existing tests, which are there to test the events and not the state of the token. Extra conditions are "hidden" in these test, when they should actually have their own test.

scheb:
Merged into 7.x and 8.x. Will go out with the next release.

Thank you for your work and your patience :)

giosh94mhz:
Thank you @scheb! 

PS: by the way, there wasn't s single line of code that passed through AI, if something feels odd maybe I just misinterpreted your instructions :smile:

scheb:
Sorry for wrongly assuming there was AI code involved, my bad! 😅

scheb:
Released as v7.14.0 / v8.6.0

PR #319, opened 2026-06-02
docs(installation): add missing link to step 3 in toc

stlgaits:
**Description**

Add missing link to install step 3 in table of contents.

image


scheb:
Thanks for the fix. Will propagate it into the other branches.

